PR #26, opened 2018-12-30
Example Gallery

brianegan:
Hey hey -- I wanted to play with hooks and get a feel for them. 

I started creating a couple of one-off examples and my own hooks to internalize the concepts and ended up creating a kind of Gallery.

I decided this might be a nice addition to the repo. It's incomplete, but I can add more examples and much more documentation if this looks like something you're interested in!

## Todo

  - [ ] Update README
  - [ ] Add documentation for each step of each example
  - [ ] Example Functional Hook
  - [ ] Example Class-based Hook
  - [ ] useMemoized
  - [ ] useStreamController & Friends
  - [ ] useAnimationController & Friends


rrousselGit:
Awesome, I love the idea!
A lot of examples is what's missing to the project.


I think it would be interesting to split this into multiple PR though. One for the gallery itself, and one per example. So that I can work on some example simultaneously with you

brianegan:
I dig it. I'll go ahead and clean up what I've got and we can use that as the starting point for the gallery.

rrousselGit:
Following this, do you think it would be interesting to add a "vanilla hooks" to Flutter architecture samples?

brianegan:
Definitely! It'd be very interesting to see how it changes the app architecture. I could see a scenario where hooks play a "lighter" role to create Blocs, Presenters, etc, or a more active Role where they contain more business logic. Since they're quite a new concept, it would be fun to play with both directions.

Anyhow, visiting family and traveling for just a couple more days! I'll clean get back to this PR on Thursday or Friday :)

rrousselGit:
Cool, I'll give it a try!

BTW I would heavily recommend using  https://github.com/rrousselGit/functional_widget in the process while working on some examples. They are meant to be used together. So that we're going from:

```dart
class Foo extends StatefulWidget {
  @override
  _FooState createState() => _FooState();
}

class _FooState extends State<Foo>
    with SingleTickerProviderStateMixin {
  AnimationController _controller;

  @override
  void initState() {
    super.initState();
    _controller = AnimationController(vsync: this);
  }

  @override
  void dispose() {
    super.dispose();
    _controller.dispose();
  }

  @override
    return Container();
  }
}
```

to:

```dart
@hwidget
Widget foo() {
  final animationController = useAnimationController();
  return Container();
}
```

brianegan:
Hey hey -- found a little bit of time to update this code, mostly small cleanup and adding documentation to help folks understand what the code is doing.

Does this look Ok as a starting point? It's a bit more filled out than what is currently available, but does not yet make use of `function_widget`.

rrousselGit:
> One recommendation: Right now the android and ios directories are ignored.

I typically use `flutter create .` for this, which bootstraps everything at once.
I'm not necessarily against it, but they don't bring any value to the example.

rrousselGit:
Well since to run the example you have to do a `flutter packages get`, but that `flutter create .` does that `packages get` internally, then it really doesn't change much.

rrousselGit:
Nice one on adding it in the readme. Obvious but I didn't think of it! 👌

brianegan:
> Well since to run the example you have to do a flutter packages get, but that flutter create . does that packages get internally, then it really doesn't change much.

Yep, it's just a `flutter run` vs a `flutter create .` + `flutter run`. IntellIJ also makes all of this automatic for ya if the android and ios folders are place, which is why I noticed it (I often just open a project and hit "Run" and everything Just Works, but didn't for this repo).

brianegan:
Looks like the build passed! When you get a chance, please let me know if this needs any further changes :)

Solido:
I've been waiting for this !

rrousselGit:
image

rrousselGit:
Nice job! It's a good starter!

We will definitely need to write some tests for these examples though (which make tests an example by themselves).

PR #80, opened 2019-05-02
ErrorDescription gone in Flutter 1.5.8

bgetsug:
I switched to the Flutter `dev` channel today and upgraded all my packages, including `flutter_bloc` which depends on this package. It seems the `ErrorDescription` class has been removed in Flutter >= 1.5.8.

### DIFF
--- a/lib/src/framework.dart
+++ b/lib/src/framework.dart
@@ -325,8 +325,7 @@ This may happen if the call to `Hook.use` is made under some condition.
             exception: exception,
             stack: stack,
             library: 'hooks library',
-            context: ErrorDescription(
-                'while calling `didBuild` on ${hook.runtimeType}'),
+            context: 'while calling `didBuild` on ${hook.runtimeType}',
           ));
         }
       }
@@ -346,7 +345,7 @@ This may happen if the call to `Hook.use` is made under some condition.
             exception: exception,
             stack: stack,
             library: 'hooks library',
-            context: ErrorDescription('while disposing ${hook.runtimeType}'),
+            context: 'while disposing ${hook.runtimeType}',
           ));
         }
       }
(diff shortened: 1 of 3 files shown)


bgetsug:
Ugh...travis failing because now it looks like the `FlutterErrorDetails,context` property is a `DiagnosticsNode` on Flutter `master`: https://github.com/flutter/flutter/blob/63aa5b3647dbd912a02f7545fc0101003cb3adc4/packages/flutter/lib/src/widgets/layout_builder.dart#L229

rrousselGit:
That's the opposite I think. `ErrorDescription` is very new, and is the reason why your PR fails with a ` The argument type 'String' can't be assigned to the parameter type 'DiagnosticsNode' • lib/src/framework.dart:328:13 • argument_type_not_assignable`


I made the switch to use `ErrorDescription` literally yesterday with https://github.com/rrousselGit/flutter_hooks/pull/79 :D

bgetsug:
Sorry, I definitely confused myself somehow. When I mentioned `flutter_bloc` above, I meant your `provider` package. I'm using `provider` 1.6.1, which has the dependency on `flutter_hooks`, though I'm not yet using hooks for anything. I've added a dependency override for `flutter_hooks` 0.3.0, which has resolved my issue.

rrousselGit:
There's soon a 2.0 release of `provider` that doesn't use `flutter_hooks` anymore anyway. You might want to check it out.

edulpn:
phew... I had some dependency that had `flutter_hooks 0.4.0` as a dependency, but had `provider 1.6.1`. It was giving me multiple `ErrorDescription` compiling errors, but worked after upgrading to `provider 2.0.0` plus minor refactors from `StatefulProvider` to `Provider`.

Issue #81, opened 2019-05-06
Dispose in reverse order


ds84182:
Currently, flutter_hooks disposes hooks from the top down, which is strange considering that Flutter itself disposes states from the bottom up. This causes exceptions when removing ChangeNotifier listeners that get cleaned up by useEffect's disposal callback.

rrousselGit:
Ah, indeed. I would swear I reversed it. I remember fixing the failing tests and everything...

Thanks, will do.

derolf:
I’ve just encountered the same problem. Can you please merge the fix?

rrousselGit:
It's not that simple. While such change fixes a few things, it also breaks others.

Similarly, React doesn't dispose of them in reverse order. And clearly, they use hooks on a much larger scale

derolf:
Could you merge it and just give the library a flag to define forward/reverse order.

rrousselGit:
@derolf  You mentioned that this caused a lot of issues. Could you clarify what these are?

For now, I'm still not convinced if we should do something about it or not.

derolf:
Generally, I think reverse order is natural since that is what you usually do in dispose/destructors: you dispose stuff in the reverse order of creation. That is especially important if there are dependencies among the things you create.

Concrete: I have a bunch of hooks to create scrollcontrollers, texteditingcontrollers etc. and then in a large fraction I am using useListenable to connect to them. During destruction, the controllers are killed first and then ListenableHook tries to remove itself which crashes the app.

rrousselGit:
The issue is, reverse order cause problems too because hooks have a "keys" that allow them to be disposed of on command.

So for example, we can have:
```dart
final controller = useMemoized(() => MyController(someKey), [someKey]);
useMyController(controller);
```

In that situation, we're stuck. Because changing `someKey` would the previous controller, but `useMyController` would still remove its listener _after_ the disposal.


So having hooks disposed _only_ from top to bottom allows some form of consistency.

rrousselGit:
Similarly, React doesn't dispose hooks in reverse order of creation, and their hooks are used on a much larger scale.


A potential solution may be to offer 3 kind of hooks instead of two (or at least 2 + a flag):
- a hook to create/dispose an object
- another one to listen to the object
- a third one that does both


Although this may not solve more advanced use-cases with animations and tweens and stuff.

derolf:
[...]

PR #86, opened 2019-05-13
reverse dispose call orders

rrousselGit:
### Feature

- [x] dispose hooks in reverse order when `HookWidget` is unmounted
- [ ] dispose hooks in reverse order when hot-reload cannot complete

### related issues:

#81


derolf:
Proposal for a smooth transition:

- Create a global static flag to specific the destruction order
- Allow each HookWidget/HookBuilder to override the flag (opt-out) by passing a flag to their constructor

rrousselGit:
Closing this since I don't have the time to work on it nor am do I have a proper solution in mind

PR #107, opened 2019-10-26
Add hook to build a text controller

### LINKED ISSUE #40: TextEditingController hook

simolus3:
This PR adds the `useTextEditingController` hook function, which can be used to construct a text editing controller with an optional initial text or `TextEditingValue`.

Fixes #40

### DIFF
--- a/lib/src/text_controller.dart
+++ b/lib/src/text_controller.dart
@@ -18,21 +18,39 @@ class _TextEditingControllerHookCreator {
   }
 }
 
-/// Functions to create a text editing controller, either via an initial
-/// text or an initial [TextEditingValue].
+/// Creates a [TextEditingController], either via an initial text or an initial
+/// [TextEditingValue].
 ///
 /// To use a [TextEditingController] with an optional initial text, use
-/// [_TextEditingControllerHookCreator.call]:
 /// ```dart
 /// final controller = useTextEditingController(text: 'initial text');
 /// ```
 ///
 /// To use a [TextEditingController] with an optional inital value, use
-/// [_TextEditingControllerHookCreator.fromValue]:
 /// ```dart
 /// final controller = useTextEditingController
 ///   .fromValue(TextEditingValue.empty);
 /// ```
+///
+/// Changing the text or initial value after the widget has been built has no
+/// effect whatsoever. To update the value in a callback, for instance after a
+/// button was pressed, use the [TextEditingController.text] or
+/// [TextEditingController.value] setters. To have the [TextEditingController]
+/// reflect changing values, you can use [useEffect]. This example will update
+/// the [TextEditingController.text] whenever a provided [ValueListenable]
+/// changes:
+/// ```dart
+/// final controller = useTextEditingController();
+/// final update = useValueListenable(myTextControllerUpdates);
+///
+/// useEffect(() {
+///   controller.text = update;
+///   return null; // we don't need to have a special dispose logic
+/// }, [update]);
+/// ```
+///
+/// See also:
+/// - [TextEditingController], which this hook creates.
 const useTextEditingController = _TextEditingControllerHookCreator();
 
 class _TextEditingControllerHook extends Hook<TextEditingController> {
@@ -49,7 +67,7 @@ class _TextEditingControllerHook extends Hook<TextEditingController> {
         super(keys: keys);
 
   @override
-  HookState<TextEditingController, _TextEditingControllerHook> createState() {
+  _TextEditingControllerHookState createState() {
     return _TextEditingControllerHookState();
   }
 }
@@ -58,20 +76,17 @@ class _TextEditingControllerHookState
     extends HookState<TextEditingController, _TextEditingControllerHook> {
   TextEditingController _controller;
 
-  TextEditingController _constructController() {
-    if (hook.initialText != null) {
-      return TextEditingController(text: hook.initialText);
-    } else if (hook.initialValue != null) {
-      return TextEditingController.fromValue(hook.initialValue);
+  @override
+  void initHook() {
+    if (hook.initialValue != null) {
+      _controller = TextEditingController.fromValue(hook.initialValue);
     } else {
-      return TextEditingController();
+      _controller = TextEditingController(text: hook.initialText);
     }
   }
 
   @override
-  TextEditingController build(BuildContext context) {
-    return _controller ??= _constructController();
-  }
+  TextEditingController build(BuildContext context) => _controller;
 
   @override
   void dispose() => _controller?.dispose();
--- a/lib/src/text_controller.dart
+++ b/lib/src/text_controller.dart
@@ -1,0 +1,59 @@
+part of 'hooks.dart';
+
+/// Creates an [TextEditingController] that will be disposed automatically.
+///
+/// The optional [initialText] parameter can be used to set the initial
+/// [TextEditingController.text]. Similarly, [initialValue] can be used to set
+/// the initial [TextEditingController.value]. It is invalid to set both
+/// [initialText] and [initialValue].
+/// When this hook is re-used with different values of [initialText] or
+/// [initialValue], the underlying [TextEditingController] will _not_ be
+/// updated. Set values on [TextEditingController.text] or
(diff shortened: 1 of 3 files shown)


rrousselGit:
Hello, thanks for your contribution!

We'll want a few changes before merging it first:
This kind of hook should strictly match the prototype of the original class.
As such, we'll want _all_ constructors parameters to be passed on the custom hook too.
Also, if a class has named constructors, the hook should follow the same pattern too.

As such, we can have something like:
```dart
useTextEditingController(initialText: 'foo');
```
and:
```dart
useTextEditingController.fromValue(value);
```

This is possible using callable classes.

simolus3:
Thanks for the quick feedback, I've changed the api to closer match the ones from Flutter by using callable classes. I'm not sure if I really like that (it negatively impacts auto-complete and generated documentation), but I understand that it's desirable to match the underlying Flutter apis as closely as possible.

Impact on coverage is because of const constructors and an assert that can no longer fail due to api misuse. Should I just remove the assert?

lib/src/text_controller.dart
```diff
@@ -0,0 +1,78 @@
...
+  TextEditingController fromValue(TextEditingValue value, [List<Object> keys]) {
+    return Hook.use(_TextEditingControllerHook(null, value, keys));
+  }
+}
+
+/// Functions to create a text editing controller, either via an initial
+/// text or an initial [TextEditingValue].
+///
+/// To use a [TextEditingController] with an optional initial text, use
+/// [_TextEditingControllerHookCreator.call]:
```
rrousselGit: The dartdoc shouldn't point to private class.

We can probably just remove the mention to [_TextEditingControllerHookCreator.call], as there's an example right next to this description.

lib/src/text_controller.dart
```diff
@@ -0,0 +1,78 @@
+part of 'hooks.dart';
+
+class _TextEditingControllerHookCreator {
+  const _TextEditingControllerHookCreator();
+
+  /// Creates a [TextEditingController] that will be disposed automatically.
+  ///
+  /// The [text] parameter can be used to set the initial value of the
+  /// controller.
```
rrousselGit: It should mention that updating `text` has no impact whatsoever.
Bonus point if you include a snippet to show how to update it (with something like `useEffect` for example)

lib/src/text_controller.dart
```diff
@@ -0,0 +1,78 @@
...
+/// ```dart
+/// final controller = useTextEditingController(text: 'initial text');
+/// ```
+///
+/// To use a [TextEditingController] with an optional inital value, use
+/// [_TextEditingControllerHookCreator.fromValue]:
+/// ```dart
+/// final controller = useTextEditingController
+///   .fromValue(TextEditingValue.empty);
+/// ```
```
rrousselGit:
```suggestion
/// ```
///
/// See also:
/// - [TextEditingController], which this hooks creates.
```

lib/src/text_controller.dart
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  /// Creates a [TextEditingController] from the initial [value] that will
+  /// be disposed automatically.
+  TextEditingController fromValue(TextEditingValue value, [List<Object> keys]) {
+    return Hook.use(_TextEditingControllerHook(null, value, keys));
+  }
+}
+
+/// Functions to create a text editing controller, either via an initial
```
rrousselGit:
```suggestion
/// Creates a [TextEditingController], either via an initial
```

test/use_text_editing_controller_test.dart
```diff
@@ -0,0 +1,57 @@
...
+        return Container();
+      },
+    ));
+
+    expect(controller.text, 'hello hooks');
+  });
+
+  testWidgets('respects initial value property', (tester) async {
+    const value = TextEditingValue(
+        text: 'foo', selection: TextSelection.collapsed(offset: 2));
```
rrousselGit:
```suggestion
        text: 'foo', selection: TextSelection.collapsed(offset: 2),
    );
```

test/use_text_editing_controller_test.dart
```diff
@@ -0,0 +1,57 @@
...
+
+    await tester.pumpWidget(HookBuilder(
+      builder: (context) {
+        controller = useTextEditingController();
+        return Container();
+      },
+    ));
+
+    expect(controller, isNotNull);
+    controller.addListener(() {});
```
rrousselGit: Thinking about that, should the hook `dispose` contains an assert that checks that the controller has no more listeners (using `controller.hasListeners`)?

test/use_text_editing_controller_test.dart
```diff
@@ -0,0 +1,57 @@
+import 'package:flutter/widgets.dart';
+import 'package:flutter_hooks/src/framework.dart';
+import 'package:flutter_hooks/src/hooks.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+import 'mock.dart';
+
+void main() {
```
rrousselGit: Both tests should try to rebuild `HookBuilder` and check that the controller is the same and that the value has not changed.

lib/src/text_controller.dart
```diff
@@ -0,0 +1,78 @@
...
+    } else if (hook.initialValue != null) {
+      return TextEditingController.fromValue(hook.initialValue);
+    } else {
+      return TextEditingController();
+    }
+  }
+
+  @override
+  TextEditingController build(BuildContext context) {
+    return _controller ??= _constructController();
```
rrousselGit: This should be moved to `initHook` for a slightly improved perf on subsequent rebuilds.
`build` will just return the controller with no extra operation.

lib/src/text_controller.dart
```diff
@@ -0,0 +1,78 @@
...
+    return _TextEditingControllerHookState();
+  }
+}
+
+class _TextEditingControllerHookState
+    extends HookState<TextEditingController, _TextEditingControllerHook> {
+  TextEditingController _controller;
+
+  TextEditingController _constructController() {
+    if (hook.initialText != null) {
```
rrousselGit: This `if` can be merged with the `else` branch

lib/src/text_controller.dart
```diff
@@ -0,0 +1,78 @@
...
+      [List<Object> keys])
+      : assert(
+          initialText == null || initialValue == null,
+          "initialText and intialValue can't both be set on a call to "
+          'useTextEditingController!',
+        ),
+        super(keys: keys);
+
+  @override
+  HookState<TextEditingController, _TextEditingControllerHook> createState() {
```
rrousselGit:
```suggestion
  _TextEditingControllerHookState createState() {
```

rrousselGit:
>  I'm not sure if I really like that (it negatively impacts auto-complete and generated documentation)

How so? This shouldn't have any impact (besides being a bit unnatural at first).

lib/src/text_controller.dart
simolus3: Right. My intention somehow was that we wouldn't have to build the controller at all if `build` never got called. That's nonsense of course.

lib/src/text_controller.dart
simolus3: I've removed both references to the private class. This also clears my concerns on callable methods with dartdoc :)

lib/src/text_controller.dart
simolus3: I've added this and an example on how to have the controller reflect changes from another `ValueListenable` to the `useTextEditingController` field, as it affects both the text and the value.

test/use_text_editing_controller_test.dart
simolus3: I'm not sure about that - I think there are valid use cases where a widget (or a builder in a subtree) listens to a text controller without calling `removeListener`. Since the widget manages the controller and all listeners, the developer can omit the call to `removeListener` without leaking resources.

After all, `_AnimationControllerHookState` doesn't do a similar check either.

rrousselGit: Okey. If it caused a problem, it's the job of ValueNotifier/ChangeNotifier to make such assert anyway

PR #123, opened 2020-01-29
add deactivate method to HookState

sahandevs:
works like AutomaticKeepAliveClientMixin


rrousselGit:
Nice!

Since you're exposing a new lifecycle on HookState, we'll want tests for that too

Also, what happens if we use that hook multiple times with different values?

It may also be worth importing all the tests for the official mixin to flutter_hooks too.

sahandevs:
thanks for the comment @rrousselGit 

1)

I added deactive to mock class but wasn't sure how to write tests for it

I can try this later it may work:
```dart
final _key = GlobalKey();
final state = ValueNotifier(false);

pumpWidget(
   ValueListenableBuilder(
     value: state,
     builder: (context, value) => Stack(
       children: [
           Container(
              child: Container(key: value ? null : _key)
           ),
          Container(
              child: Container(key: !value ? null : _key)
           )
       ]
    )
   )
)
pump();
expect(deactivate, noCall);
state.value = true;
pump();
expect(deactivate, calledOne);
pump();
verifyNoMoreInteractions(deactivate);
```

2)
In case of `wantKeepAlive:false` it doesn't do anything and in case of true it calls `KeepAliveNotification(_keepAliveHandle).dispatch(context);` (which calls `visitAncestorElements`)multiple times.
It doesn't cause any weird behaviors but calling `visitAncestorElements` is expensive so it isn't really ideal.
we can like `useSingleTickerProvider` notify user (only in debug mode) that you should use this hook only once.
```dart
void useAutomaticKeepAliveClient({bool wantKeepAlive = true}) {
  assert(() { 
      final context = useContext();
      final haveMultipleAutomaticKeepAliveClientHooks = _hooks.where((item) => item is _AutomaticKeepAliveClientHookState).length > 1;
      return  !haveMultipleAutomaticKeepAliveClientHooks ;  }());
  Hook.use(_AutomaticKeepAliveClientHook(wantKeepAlive));
}
```
to do above we have to make _hooks public (or better, make them protected)

3)
sure I will add them

rrousselGit:
> I added deactive to mock class but wasn't sure how to write tests for it
> I can try this later it may work:

Yes that should be fine :)

> It doesn't cause any weird behaviors but calling visitAncestorElements is expensive so it isn't really ideal.

If we do:

```dart
useAutomaticKeepAliveClient(true);
useAutomaticKeepAliveClient(false);
```

What should/would happen?

It may be more logical to implement it as a widget instead of a hook:


```dart
Widget build(context) {
  final foo = useWhatever();

  return AutomaticKeepAlive(
    keepAlive: true,
    child: Whatever(),
  );
}
```

This will prevent such weird unknown scenario from happening, and we wouldn't have to fork the implementation of the mixin.\
It may also be worth trying to merge such a widget in Flutter directly – it should be doable.

What do you think?

sahandevs:
I searches a little bit and it seems there are two widgets, `AutomaticKeepAlive` (parent) and `KeepAlive` (child) ,that do exactly that. I think you are right `useAutomaticKeepAliveClient` may be unnecessary in most of the situations.
but i think there are some useful cases.

example: changing the tab in the main menu of an app should not destroy state of the other tabs (for example scroll position)

```dart
TabController useMainRouteTabs({ bool keepChildAlive = true }) {
  useAutomaticKeepAliveClient(keepAlive: keepChildAlive );
  return useMemo(() => TabController(), []);
}
```

of course there are other solutions to the problem above but by writing just one line of code (`useMainRouteTabs();`) you have all the logic you need, you don't need to wrap your child widgets with another widget to keep the state. It's just my opinion but I think grouping __all__ the logic you need in a single function call is the thing that makes hooks powerful.

what about removing the `useAutomaticKeepAliveClient` and keeping the `deactivate` in `HookState` and let the users decide?

rrousselGit:
> what about removing the useAutomaticKeepAliveClient and keeping the deactivate in HookState and let the users decide?

Definitely. That sounds more healthy for now. If people keep asking for such hook, we can reconsider 😄

deadsoul44:
I need this hook to use in may app where I have several tab views and most of the tabs are built with data coming from API. I make API calls on tab change. But, I want to keep tabs alive. So that, I don't have to fetch data again and I don't have to show ugly CircularProgressIndicator again and again unnecessarily. What is your suggestion? I don't want to move back to StatefulWidget.

sahandevs:
@rrousselGit I removed the hooks and added test for deactivate

@deadsoul44 
if this PR merges you can use this code in your project

```dart
/// Allows subtrees to request to be kept alive in lazy lists.
///
/// See also:
///
///  * [AutomaticKeepAlive]
///  * [AutomaticKeepAliveClientMixin]
void useAutomaticKeepAliveClient({bool wantKeepAlive = true}) {
  Hook.use(_AutomaticKeepAliveClientHook(wantKeepAlive));
}

class _AutomaticKeepAliveClientHook extends Hook<void> {
  final bool wantKeepAlive;

  _AutomaticKeepAliveClientHook(this.wantKeepAlive);

  @override
  _AutomaticKeepAliveClientHookState createState() =>
      _AutomaticKeepAliveClientHookState();
}

class _AutomaticKeepAliveClientHookState
    extends HookState<void, _AutomaticKeepAliveClientHook> {
  KeepAliveHandle _keepAliveHandle;

  bool get wantKeepAlive => hook.wantKeepAlive;

  void _ensureKeepAlive() {
    assert(_keepAliveHandle == null);
    _keepAliveHandle = KeepAliveHandle();
    KeepAliveNotification(_keepAliveHandle).dispatch(context);
  }

  void _releaseKeepAlive() {
    _keepAliveHandle.release();
    _keepAliveHandle = null;
  }

  /// Ensures that any [AutomaticKeepAlive] ancestors are in a good state, by
  /// firing a [KeepAliveNotification] or triggering the [KeepAliveHandle] as
  /// appropriate.
  @protected
  void updateKeepAlive() {
    if (wantKeepAlive) {
      if (_keepAliveHandle == null) _ensureKeepAlive();
    } else {
      if (_keepAliveHandle != null) _releaseKeepAlive();
    }
  }

  @override
  initHook() {
    super.initHook();
    if (wantKeepAlive) _ensureKeepAlive();
  }

  @override
  void deactivate() {
    if (_keepAliveHandle != null) _releaseKeepAlive();
    super.deactivate();
  }

  @override
  void build(BuildContext context) {
    if (wantKeepAlive && _keepAliveHandle == null) _ensureKeepAlive();
  }
}
```

test/hook_widget_test.dart
```diff
@@ -50,11 +52,50 @@ void main() {
...
+                  ]
+              )
+          ),
+        )
+    );
+    await tester.pump();
+    verifyNever(deactivate());
+    state.value = true;
+    await tester.pump();
+    verify(deactivate()).called(1);
```
rrousselGit: Could you expand it slightly to assert that it calls deactivate on all hooks in order?

And another check that if one deactivate call fails, the others are still called

rrousselGit: Also it may actually not be needed to use GlobalKey here, since `deactivate` is also called right before `dispose`

sahandevs:
@rrousselGit 
I can't get it to work, can you see whats wrong?

deadsoul44:
I tried AutomaticKeepAlive and KeepAlive widgets but I get the following error:

image

Any other suggestion other than using the useAutomaticKeepAliveClient code?

Thank you.

rrousselGit:
> I can't get it to work, can you see whats wrong?

What do you mean by this?

sahandevs:
@rrousselGit I added the two test you menitioned, but i can't figure out why `should call other deactivates even if one fails` doesn't work

rrousselGit:
What do you mean by "it doesn't work"?

sahandevs:
It doesn't throw an error.

rrousselGit:
Sorry for the late reply

That's because you used `expectPump`. You should probably just use `tester.pumpWidget` and overload `FlutterError.onError`.

This would also fill the uncovered lines

sahandevs:
@rrousselGit I think it's ok now.
the problem was flutter tester itself was calling `deactivate` after the test and it complained about the exception deactivate throws! fixed by resetting `when(deactivate)... `

KeithMarcusXIII:
> @rrousselGit I removed the hooks and added test for deactivate
> 
> @deadsoul44
> you can already use https://api.flutter.dev/flutter/widgets/AutomaticKeepAlive/AutomaticKeepAlive.html and https://api.flutter.dev/flutter/widgets/KeepAlive-class.html without the need of StatefulWidget
> 
> or if you want the hook version, if this PR merges you can use following code in your project
> 
> ```dart
> /// Allows subtrees to request to be kept alive in lazy lists.
> ///
> /// See also:
> ///
> ///  * [AutomaticKeepAlive]
> ///  * [AutomaticKeepAliveClientMixin]
> void useAutomaticKeepAliveClient({bool wantKeepAlive = true}) {
>   Hook.use(_AutomaticKeepAliveClientHook(wantKeepAlive));
> }
> 
> class _AutomaticKeepAliveClientHook extends Hook<void> {
>   final bool wantKeepAlive;
> 
>   _AutomaticKeepAliveClientHook(this.wantKeepAlive);
> 
>   @override
>   _AutomaticKeepAliveClientHookState createState() =>
>       _AutomaticKeepAliveClientHookState();
> }
> 
> class _AutomaticKeepAliveClientHookState
>     extends HookState<void, _AutomaticKeepAliveClientHook> {
>   KeepAliveHandle _keepAliveHandle;
> 
>   bool get wantKeepAlive => hook.wantKeepAlive;
> 
>   void _ensureKeepAlive() {
>     assert(_keepAliveHandle == null);
>     _keepAliveHandle = KeepAliveHandle();
>     KeepAliveNotification(_keepAliveHandle).dispatch(context);
>   }
> 
>   void _releaseKeepAlive() {
>     _keepAliveHandle.release();
>     _keepAliveHandle = null;
>   }
> 
>   /// Ensures that any [AutomaticKeepAlive] ancestors are in a good state, by
>   /// firing a [KeepAliveNotification] or triggering the [KeepAliveHandle] as
>   /// appropriate.
>   @protected
>   void updateKeepAlive() {
>     if (wantKeepAlive) {
>       if (_keepAliveHandle == null) _ensureKeepAlive();
>     } else {
>       if (_keepAliveHandle != null) _releaseKeepAlive();
>     }
>   }
> 
>   @override
>   initHook() {
>     super.initHook();
>     if (wantKeepAlive) _ensureKeepAlive();
>   }
> 
>   @override
>   void deactivate() {
>     if (_keepAliveHandle != null) _releaseKeepAlive();
>     super.deactivate();
>   }
> 
>   @override
>   void build(BuildContext context) {
>     if (wantKeepAlive && _keepAliveHandle == null) _ensureKeepAlive();
>   }
> }
> ```

I think this hook maybe needed. If I use KeepAlive and AutomaticKeepAlive on HookWidget, it will complain 
> Incorrect use of ParentDataWidget.
And I was using this on children of PageView and it wasn't working. However I switched to the hook and it worked

PR #128, opened 2020-03-30
Fix Existing hooks link on Pub

creativecreatorormaybenot:
If you click on the current link in that section on Pub, it will *not* work. It will direct you to `https://github.com/rrousselGit/flutter_hooks#existing-hooks)` instead of `https://github.com/rrousselGit/flutter_hooks#existing-hooks`, which will not *jump* to the correct heading.  
On GitHub it works, but [on the Pub page](https://pub.dev/packages/flutter_hooks#motivation), it does not work.

I took this opportunity to shorten the link by giving it an alias.  
You could instead also link to it like this: `[https://github.com/rrousselGit/flutter_hooks#existing-hooks](https://github.com/rrousselGit/flutter_hooks#existing-hooks)`. This would both fix the link and retain the old look, however, I do not think that this is desirable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.8.0+1
+
+- Fixed link to "Existing hooks" in `README.md`.
+
 ## 0.8.0:
 
 Added `useFocusNode`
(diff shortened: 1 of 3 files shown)


rrousselGit:
Thanks!

PR #151, opened 2020-07-06
Add useTabController hook

Albert221:
Hi! I just stumbled upon a case where I wanted to use a `TabController` with flutter_hooks, but it quickly came out that I can't use it in `useMemoized` because it's not disposing the controller properly (well, it does not dispose it at all). So I created a new hook just for the `TabController`.

### Sample usage
      final ticker = useSingleTickerProvider();
      final tabController = useTabController(length: 2, vsync: ticker);

      return Scaffold(
        appBar: AppBar(
          title: ...,
          bottom: TabBar(
            controller: tabController,
            tabs: [
              // ...


lib/src/tab_controller.dart
```diff
@@ -0,0 +1,56 @@
+part of 'hooks.dart';
+
+/// Creates and disposes a [TabController].
+///
+/// See also:
+/// - [TabController]
+TabController useTabController({
```
VarmaMSP, line 7: Keeping vsync optional here would be better.

Albert221: Good point! I fixed that

rrousselGit:
Thanks!

PR #162, opened 2020-08-01
useIsMounted hook

### LINKED ISSUE #138: Having a mounted properties when using useState
One user of my package report me this https://github.com/mylisabox/flutter_mjpeg/issues/6 
Look like the ValueNotifier returned by useState is disposed by the time I want to change it. Having a `mounted` property would allow me to check if it's disposed or not before trying to modify it.

davidmartos96:
Resolves #138


satvikpendem:
Perhaps the name `useMounted` is better as other boolean hooks don't use the `is` word.

davidmartos96:
@satvikpendem Good point! I opted for the isMounted because I think it makes more clear that is a callable class. 

```dart
final isMounted = useIsMounted();
print(isMounted());
```
What do you think @rrousselGit ?

lib/src/hooks.dart
```diff
@@ -16,3 +16,4 @@ part 'tab_controller.dart';
 part 'text_controller.dart';
 part 'focus.dart';
 part 'scroll_controller.dart';
+part 'is_mounted.dart';
```
rrousselGit, line 19: Place it inside mist instead

lib/src/is_mounted.dart
```diff
@@ -0,0 +1,38 @@
+part of 'hooks.dart';
+
+/// Returns an [IsMounted] object that you can use
+/// to check if the [State] is mounted.
+/// ```dart
```
rrousselGit, line 5: Nit: we need a newline before ```dart

lib/src/is_mounted.dart
```diff
@@ -0,0 +1,38 @@
...
+///     if (isMounted()) {
+///       // Do something
+///     }
+///   });
+///   return null;
+/// }, []);
+/// ```
+/// See also:
+///   * The [State.mounted] property.
+IsMounted useIsMounted() {
```
rrousselGit, line 18: Could you use the class syntax instead?
Internal hooks all use classes for better hot-reload

rrousselGit:
useIsMounted sounds alright to me, as the class does not return a boolean.

lib/src/is_mounted.dart
davidmartos96: Oh, I didn't know that. Sure, I'll change it!

lib/src/hooks.dart
davidmartos96: Done!

davidmartos96:
@rrousselGit The codecov is not liking the "formatting". Do I have to squash and force push?

rrousselGit:
Don't worry about it. Thanks!

Afsar-Pasha:
What about something like this, it doesn't requires new object

```dart
typedef IsMounted = bool Function();

class _IsMountedHookState extends HookState<IsMounted, _IsMountedHook> {
  bool mounted = true;

  bool isMounted() => mounted;

  @override
  IsMounted build(BuildContext context) => isMounted;

  @override
  void dispose() {
    mounted = false;
    super.dispose();
  }
}
```

davidmartos96:
> What about something like this, it doesn't requires new object
> 
> ```dart
> typedef IsMounted = bool Function();
> 
> class _IsMountedHookState extends HookState<IsMounted, _IsMountedHook> {
>   bool mounted = true;
> 
>   bool isMounted() => mounted;
> 
>   @override
>   IsMounted build(BuildContext context) => isMounted;
> 
>   @override
>   void dispose() {
>     mounted = false;
>     super.dispose();
>   }
> }
> ```

That's a nice idea. When I submitted the PR I haven't though about returning a function. It may be clearer than using the callable class.

PR #167, opened 2020-08-14
Add support for using RestorableProperty with hook

blaugold:
This PR adds support for managing `RestorableProperty`s with the help of a hook and `HookRestorationScope`.

### DIFF
--- a/example/lib/use_restorable_property.dart
+++ b/example/lib/use_restorable_property.dart
@@ -1,0 +1,46 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_hooks/flutter_hooks.dart';
+
+/// This example demonstrate what is required for and how to use
+/// [useRestorationProperty].
+class RestorablePropertyExample extends StatelessWidget {
+  @override
+  Widget build(BuildContext context) {
+    return Scaffold(
+      appBar: AppBar(
+        title: const Text('useRestorationProperty example'),
+      ),
+      // Any app using the restoration system needs to make the root
+      // `RestorationBucket` available to the widget tree. Usually this widget
+      // is placed high up in the tree, for example above `MaterialApp`.
+      body: RootRestorationScope(
+        restorationId: 'root',
+        // This widget provides hooks access to a `RestorationBucket`. It is a
+        // direct replacement for `RestorationScope` from the framework.
+        child: HookRestorationScope(
+          restorationId: 'hooks',
+          child: HookBuilder(
+            builder: (context) {
+              // Here we register a restorable integer counter under the
+              // `restorationId` `count`. The property passed to the hook during
+              // the first build is the only value being used. Every time
+              // `useRestorationProperty` is called the same value is returned.
+              // In addition to registering the property, the hook also listens
+              // to it, so that changing it triggers a rebuild.
+              final count = useRestorationProperty('count', RestorableInt(0));
+              return Center(
+                child: ElevatedButton(
+                  // To update the property we simply assign a new value, which
+                  // triggers a rebuild and also makes sure the new value is
+                  // persisted by the restoration system.
+                  onPressed: () => count.value += 1,
+                  child: Text('${count.value} clicks'),
+                ),
+              );
+            },
+          ),
+        ),
+      ),
+    );
+  }
+}
--- a/lib/src/restoration.dart
+++ b/lib/src/restoration.dart
@@ -1,0 +1,241 @@
+part of 'hooks.dart';
+
+/// Widget which behaves like [RestorationScope] but additional enables the use
+/// of [useRestorationProperty] in its children.
+class HookRestorationScope extends StatefulWidget {
+  /// Creates a [HookRestorationScope].
+  ///
+  /// Providing null as the [restorationId] turns off state restoration for
+  /// the [child] and its descendants.
+  ///
+  /// The [child] must not be null.
+  const HookRestorationScope({
+    Key key,
+    @required this.restorationId,
+    @required this.child,
+  })  : assert(child != null, 'child cannot be null.'),
+        super(key: key);
+
+  /// The widget below this widget in the tree.
+  ///
+  /// {@macro flutter.widgets.child}
+  final Widget child;
+
+  /// The restoration ID used by this widget to obtain a child bucket from the
+  /// surrounding [RestorationScope].
+  ///
+  /// The child bucket obtained from the surrounding scope is made available to
+  /// descendant widgets via [RestorationScope.of].
+  ///
+  /// If this is null, [RestorationScope.of] invoked by descendants will return
+  /// null which effectively turns off state restoration for this subtree.
+  final String restorationId;
+
+  @override
+  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
+    super.debugFillProperties(properties);
+    properties.add(StringProperty('restorationId', restorationId));
+  }
+
+  @override
+  _HookRestorationScopeState createState() => _HookRestorationScopeState();
+}
+
+/// A [State] implementation which allows registration of [RestorableProperty]s
+/// from a [HookState].
+class _HookRestorationScopeState extends State<HookRestorationScope>
+    with RestorationMixin {
+  /// Currently registered [RestorableProperty]s by restorationId.
+  final _properties = <String, RestorableProperty<void>>{};
+
+  @override
+  Widget build(BuildContext context) {
+    return _HookRestorationScopeMarker(
+      scope: this,
+      child: UnmanagedRestorationScope(
+        bucket: bucket,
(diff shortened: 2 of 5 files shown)


example/lib/use_restorable_property.dart
```diff
@@ -0,0 +1,46 @@
...
+        title: const Text('useRestorationProperty example'),
+      ),
+      // Any app using the restoration system needs to make the root
+      // `RestorationBucket` available to the widget tree. Usually this widget
+      // is placed high up in the tree, for example above `MaterialApp`.
+      body: RootRestorationScope(
+        restorationId: 'root',
+        // This widget provides hooks access to a `RestorationBucket`. It is a
+        // direct replacement for `RestorationScope` from the framework.
+        child: HookRestorationScope(
```
rrousselGit, line 20: We shouldn't need such widget. Relying on an InheritedWidget to work around the `RestorationMixin` could be a significant source of problems.

I would expect people to easily forget to wrap a custom `HookWidget` using `useRestorableProperty` into a `HookRestorationScope` – and in fact, they shouldn't have to. 

`useRestorableProperty` should be able to handle everything by itself

lib/src/restoration.dart
```diff
@@ -0,0 +1,241 @@
...
+  );
+
+  final result = use(_RestorablePropertyHook(
+    restorationId: restorationId,
+    property: property,
+    scope: scope,
+  ));
+
+  if (listen) {
+    useListenable(result);
```
rrousselGit, line 237: Conditionally calling hooks is anti-pattern. The `listen` flag probably shouldn't be there at all

example/lib/use_restorable_property.dart
blaugold: I agree, it's not optimal. The reason I implemented it this way is that the implementation of `RestorableProperty` is tied to `RestorationMixin`. I have not looked into if it would be possible to use `RestorableProperty` without `RestorationMixin` but that doesn't seem like a good idea. If users can't use `RestorableProperty`, they also can't use all the subclasses which would mean quite a large part of the user facing restoration API would need to be replaced somehow. I'll take a second look at how to get around `HookRestorationScope`.

lib/src/restoration.dart
blaugold: I assumed most users will want to listen to the property changes but thinking about it I'm not so sure. I'll remove the option and document that properties need to be listened to explicitly for example with `useListenable`.

rrousselGit: Sorry, I meant "always listen" instead, rather than optionally listen.

If we want to support optional listening, this should probably be in a different hook rather than as a flag

example/lib/use_restorable_property.dart
rrousselGit: The problem is, we can't reasonably expose the `RestorationMixin` through an InheritedWidget

That could cause major issues, as the mixin could suddenly be associated with multiple `HookWidget`

rrousselGit:
I think there are a few paths we can explore for the `RestorationMixin` issue.

In the end, the only thing we care about here is passing the `restorationId`. Then everything else can be managed by hooks or HookElement.

So we have a few options:

- A getter on the wiget:
```dart
class Example extends HookWidget {
  @override
  String get restorationId => 'foo';
}
```

- A custom hook
```dart
class Example extends HookWidget {
  @override
  Widget build(context) {
    useRestoration('foo');
    final counter = useRestorableInt(0);
  }
}
```





I also believe that we shouldn't have a single `useRestorableProperty` but rather multiple variants of `useState` that are restorable, so:

```dart
final count = useRestorableInt(0, 'count');
final str = useRestorableString('', 'str');
...
```

TimDepping:
> I think there are a few paths we can explore for the `RestorationMixin` issue.
> 
> In the end, the only thing we care about here is passing the `restorationId`. Then everything else can be managed by hooks or HookElement.
> 
> So we have a few options:
> 
> * A getter on the wiget:
> 
> ```dart
> class Example extends HookWidget {
>   @override
>   String get restorationId => 'foo';
> }
> ```
> 
> * A custom hook
> 
> ```dart
> class Example extends HookWidget {
>   @override
>   Widget build(context) {
>     useRestoration('foo');
>     final counter = useRestorableInt(0);
>   }
> }
> ```
> 
> I also believe that we shouldn't have a single `useRestorableProperty` but rather multiple variants of `useState` that are restorable, so:
> 
> ```dart
> final count = useRestorableInt(0, 'count');
> final str = useRestorableString('', 'str');
> ...
> ```

Could you share your thoughts on how to face the `RestorationMixin` issue?

blaugold:
I have incorporated the comments from @rrousselGit in this API proposal:

## Enabling restoration for hook widget

```dart 
RestorationBucket? useRestoration(String? restorationId);
```

This hook enables restoration for a hook widget. It can be used only once in a hook widget. It must have been used before other restoration hooks are used.

It takes the `restorationId` which should be used for the widget. `null` is accepted to signal that restoration is disabled for the widget (similar to how `RestorationMixin` works).

It returns the `RestorationBucket` associated with the widget if one is available. This is useful to expose a widget's bucket to its children, through `UnmanagedRestorationScope`.

## Using a `RestorableProperty`

```dart
T useRestorableProperty<T extends RestorableProperty<dynamic>>(
  ValueGetter<T> createProperty, 
  String restorationId,
);
```

This hook takes a factory, creates the property during initialization and memorizes it. It also takes the `restorationId` for the property and uses this id to register the property for restoration.

The value of the hook is the memorized property. The hook listens to the property and rebuilds when it changes.

## Hooks for existing `RestorableProperty` implementations

For existing implementations of `RestorableProperty` hooks similar to this one are provided:

```dart
RestorableInt useRestorableInt(int defaultValue, String restorationId) {
  return useRestorableProperty(
    () => RestorableInt(defaultValue),
    restorationId,
  );
}
```

## Implementation

To facilitate the use of `RestorableProperty`, `HookElement` manages an internal widget which serves as the hosts for a `RestorationMixin`.

rrousselGit:
> To facilitate the use of RestorableProperty, HookElement manages an internal widget which serves as the hosts for a RestorationMixin.

What do you mean by this?

blaugold:
We need a `RestorationMixin` somewhere to register `RestorableProperty`s with. The idea is to create a `StatefulWidget`, whose `State` mixes in `RestorationMixin` and just passes though a child. When restoration is enabled for a widget, `HookElement` inserts this widget above the widget built by it. This widget provides APIs for setting the `restorationId`, accessing the`RestorationBucket` it is managing and registering properties for restoration.

rrousselGit:
We don't need the `RestorationMixin`, but rather its implementation

It's a story similar to `SingleTickerProviderStateMixin`. flutter_hooks does not use the official mixin, but rather copy-pasted its implementation and converted it in a hook.

We could do the same thing here, and have `useRestorationProperty` host the implementation

blaugold:
If we want to avoid `RestorationMixin` we cannot use `RestorableProperty` because its implementation is tied to `RestorationMixin`. That means we will need our own `HookRestorableProperty`, which is what I wanted to avoid, but I'm OK with going that route. The proposed API would not change except `RestorableProperty` becomes `HookRestorableProperty`.

rrousselGit:
We can try an `implements RestorationMixin` to see if that is feasible. If not, making a PR to Flutter to separate `RestorableProperty` from `RestorationMixin` is a possibility.

blaugold:
Looking at [`RestorationProperty._register`](https://github.com/flutter/flutter/blob/127e67902e8bbb0dcbfb3351b8fd00f7cbdf0178/packages/flutter/lib/src/widgets/restoration.dart#L488) I doubt implementing `RestorationMixin` will work. `RestorationMixin` calls private methods of `RestorableProperty` to properly register and unregister it.

jaumard:
> We can try an implements RestorationMixin to see if that is feasible. If not, making a PR to Flutter to separate RestorableProperty from RestorationMixin is a possibility.
Looking at RestorationProperty._register I doubt implementing RestorationMixin will work. RestorationMixin calls private methods of RestorableProperty to properly register and unregister it.

Might worth asking flutter team to do that then by just make the private method public it could be solved

blaugold:
I don't think there is a good way to build on `RestorationProperty` and `RestorationMixin` 
- `RestorationMixin` is a mixin on `State`
- `RestorationProperty` has a getter `state`, which returns the `RestorationMixin` it is registered with
- Anything implementing `RestorationMixin` also has to implement `State`
- The element of `HookWidget` cannot implement `State.widget` because its `HookWidget` is a `StatelessWidget` and `State.widget` has to be  `StatefullWidget`

blaugold:
I'm closing this PR since it seems to me that implementing this properly is more trouble than it's worth.

PR #175, opened 2020-08-25
Adding full test coverage for useTextEdittingController

kranfix:
Adding full test coverage for useTextEdittingController.


lib/src/text_controller.dart
```diff
@@ -51,18 +51,20 @@ class _TextEditingControllerHookCreator {
 ///
 /// See also:
 /// - [TextEditingController], which this hook creates.
-const useTextEditingController = _TextEditingControllerHookCreator();
```
rrousselGit, line 54: Const is important. It is a tiny bit faster than using final. It's negligible, but code-coverage for this is negligible too.

rrousselGit:
Thanks!

kranfix:
@rrousselGit I revet that line to const.

PR #176, opened 2020-08-25
Generalizing useFocusNode with input parameters

kranfix:
`FocusNode` constructor has 5 params:

```dart
FocusNode({
  String debugLabel,
  FocusOnKeyCallback onKey,
  bool skipTraversal = false,
  bool canRequestFocus = true,
  bool descendantsAreFocusable = true,
})	
```

Then, they should be inputs for `useFocusNode`.


lib/src/focus.dart
```diff
@@ -16,7 +41,18 @@ class _FocusNodeHook extends Hook<FocusNode> {
...
 class _FocusNodeHookState extends HookState<FocusNode, _FocusNodeHook> {
-  final _focusNode = FocusNode();
+  FocusNode _focusNode;
+
+  @override
+  void initHook() {
+    _focusNode = FocusNode(
+      debugLabel: hook.debugLabel,
+      onKey: hook.onKey,
+      skipTraversal: hook.skipTraversal,
```
rrousselGit, line 51: All of these can change over time, so we need to handle the update scenario

rrousselGit:
We'll need a test to make sure that the default values used by the hook match those of FocusNode

rrousselGit:
I've added the tests

Thanks!

PR #177, opened 2020-08-25
Use scroll controller test

kranfix:
The file `use_scroll_controller.dart` must be renamed to `use_scroll_controller_test.dart` to include in test coverage.


rrousselGit:
Nice catch 🤦

PR #178, opened 2020-08-25
HookWidget.debugFillProperties using HookState.diagnosticsProperty

kranfix:
In the package example, the widget inspector for the useState example shows nothing about the hook state:

With this new feature, if these lines are added to the `_StateHookState`, this will be shown int the widget inspector.

```dart
class _StateHookState<T> extends HookState<ValueNotifier<T>, _StateHook<T>> {
  ValueNotifier<T> _state;

 ///...

  @override
  DiagnosticsProperty get diagnosticsProperty {
    return DiagnosticsProperty<T>('useState<$T>', _state.value);
  }
}
```


lib/src/framework.dart
```diff
@@ -281,6 +281,12 @@ abstract class HookState<R, T extends Hook<R>> {
       .._isOptionalRebuild = false
       ..markNeedsBuild();
   }
+
+  /// The [diagnosticsProperty] is the default propertity to add to `properties`
+  /// param of the [Element.debugFillProperties].
+  /// If it's null, it will be skipped
+  @visibleForTesting
+  DiagnosticsProperty get diagnosticsProperty => null;
```
rrousselGit, line 289: I think it would be more logical to use same pattern as everywhere else in Flutter: `void debugFillProperties(properties)`

rrousselGit:
Thanks!

We will need some tests for this.

lib/src/framework.dart
rrousselGit: It would probably make sense to have `HookState` implement `Diagnosticable`

kranfix:
According to the coverage/lcov.info, the `HookElement.debugFillProperties` is covered. Any suggestion for adding a local tests for this feature?

lib/src/framework.dart
```diff
@@ -203,7 +203,7 @@ Calling them outside of build method leads to an unstable state and is therefore
 }
 
 /// The logic and internal state for a [HookWidget]
-abstract class HookState<R, T extends Hook<R>> {
+abstract class HookState<R, T extends Hook<R>> with Diagnosticable {
```
rrousselGit, line 206: Hook probably should be Diagnosticable too.

kranfix: What's the need to make `Hook` diagnosticable?

rrousselGit:
Covered doesn't mean tested. We need a test that verifies that `debugFillProperties` works correctly.

A good start may be to have some existing hooks override debugFillProperties, like useState/useMemoized, and test that it works properly.

kranfix:
I don't find a way to test the debugFillProperties. Do you have any suggestion?

kranfix:
@rrousselGit I found a way to test de `debugFillProperties` . I'll be sending new commits on weekend.

kranfix:
Tests for useState and useMemoized where added

rrousselGit:
I'll do that tomorrow.
I plan to add Diagnistics for all hooks at once.

And the Diagnostics must be updated such that we can have:

```
HookBuilder
 │ useMemoized<int>: 43
 | useAnimationController: AnimationController#00000
 | └duration: Duration(seconds: 2)
└SizedBox(renderObject: RenderConstrainedBox#00000)
```

PR #182, opened 2020-08-31
Use reducer example

Sahtair:
This is a basic example of the counter implemented with a useReducer hook. It might be a bit over kill to use a reducer for this example but I think i would help many people get a grasp of the concept of a reducer hook.


sesam:
🎉 ! I think this should be enough to get me started on learning to useReducer.

rrousselGit:
Nice, thanks a lot!

Issue #190, opened 2020-10-03
Gesture recognizers hooks


Albert221:
Gesture recognizers need to be disposed. Hooks for them would make using them correctly really simple.

I'm creating this issue to track which gesture recognizers still need a hook and which don't. I'm gonna start working on the Pull Requests.

- [`GestureRecognizer`](https://api.flutter.dev/flutter/gestures/GestureRecognizer-class.html)
  - [ ] [`DoubleTapGestureRecognizer`](https://api.flutter.dev/flutter/gestures/DoubleTapGestureRecognizer-class.html)
  - [`MultiDragGestureRecognizer`](https://api.flutter.dev/flutter/gestures/MultiDragGestureRecognizer-class.html)
    - [ ] [`DelayedMultiDragGestureRecognizer`](https://api.flutter.dev/flutter/gestures/DelayedMultiDragGestureRecognizer-class.html)
    - [ ] [`HorizontalMultiDragGestureRecognizer`](https://api.flutter.dev/flutter/gestures/HorizontalMultiDragGestureRecognizer-class.html)
    - [ ] [`ImmediateMultiDragGestureRecognizer`](https://api.flutter.dev/flutter/gestures/ImmediateMultiDragGestureRecognizer-class.html)
    - [ ] [`VerticalMultiDragGestureRecognizer`](https://api.flutter.dev/flutter/gestures/VerticalMultiDragGestureRecognizer-class.html)
  - [ ] [`MultiTapGestureRecognizer`](https://api.flutter.dev/flutter/gestures/MultiTapGestureRecognizer-class.html)
  - [`OneSequenceGestureRecognizer`](https://api.flutter.dev/flutter/gestures/OneSequenceGestureRecognizer-class.html)
    - [`DragGestureRecognizer`](https://api.flutter.dev/flutter/gestures/DragGestureRecognizer-class.html)
      - [ ] [`HorizontalDragGestureRecognizer`](https://api.flutter.dev/flutter/gestures/HorizontalDragGestureRecognizer-class.html)
      - [ ] [`PanGestureRecognizer`](https://api.flutter.dev/flutter/gestures/PanGestureRecognizer-class.html)
      - [ ] [`VerticalDragGestureRecognizer`](https://api.flutter.dev/flutter/gestures/VerticalDragGestureRecognizer-class.html)
    - [ ] [`EagerGestureRecognizer`](https://api.flutter.dev/flutter/gestures/EagerGestureRecognizer-class.html)
[...]

PR #191, opened 2020-10-03
Generic GestureRecognizer and TapGestureRecognizer hooks

Albert221:
See #190.


lib/src/gesture_recognizers.dart
```diff
@@ -0,0 +1,55 @@
+part of 'hooks.dart';
+
+abstract class _GestureRecognizerHook<G extends GestureRecognizer>
```
Albert221, line 3: Generic hook class is to make creating gesture detectors hooks involve as small code duplication as possible.

example/.gitignore
```diff
@@ -23,6 +23,7 @@
 **/doc/api/
 .dart_tool/
 .flutter-plugins
+.flutter-plugins-dependencies
```
Albert221, line 26: I ignored this file as it should be ignored! :)

lib/src/hooks.dart
```diff
@@ -9,10 +10,11 @@ import 'framework.dart';
 
 part 'animation.dart';
 part 'async.dart';
+part 'focus.dart';
```
Albert221, line 13: I sorted those `part` directives alphabetically.

test/gesture_detectors_test.dart
```diff
@@ -0,0 +1,69 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/gestures.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:flutter_hooks/src/framework.dart';
+import 'package:flutter_hooks/src/hooks.dart';
+
+import 'mock.dart';
+
+void _testGestureRecognizer<T>({
```
Albert221, line 10: Here is also one generic method that makes sure hook works as hooks should for all gesture detectors that would be tested here.

rrousselGit: I don't like the idea of having such abstraction.
I'd prefer removing it and make such that the expectations don't depend on anything

test/gesture_detectors_test.dart
```diff
@@ -0,0 +1,69 @@
...
+  Type gestureRecognizer,
+  T Function() useHook,
+}) {
+  group('use$gestureRecognizer', () {
```
rrousselGit, line 14: Could we add a test where the gesture detector is used, with a click & all?

test/gesture_detectors_test.dart
Albert221: Done!

test/gesture_detectors_test.dart
Albert221: Done!

rrousselGit:
Closing as per https://github.com/rrousselGit/flutter_hooks/issues/190#issuecomment-703264898

I appreciate your work, but I just don't think that is the correct path for this kind of thing

PR #209, opened 2020-12-30
Migrate all use* methods to extension methods on Hookable

noinskit:
This prevents runtime errors when use* methods are mistakenly called in
e.g. StatelessWidget (which can and has happened in production due to
otherwise innocent refactoring), instead catching them at compile time.

### DIFF
--- a/example/lib/custom_hook_function.dart
+++ b/example/lib/custom_hook_function.dart
@@ -31,17 +31,19 @@ class CustomHookFunctionExample extends HookWidget {
   }
 }
 
-/// A custom hook that wraps the useState hook to add logging. Hooks can be
-/// composed -- meaning you can use hooks within hooks!
-ValueNotifier<T> useLoggedState<T>([T initialData]) {
-  // First, call the useState hook. It will create a ValueNotifier for you that
-  // rebuilds the Widget whenever the value changes.
-  final result = useState<T>(initialData);
+extension UseLoggedStateHook on Hookable {
+  /// A custom hook that wraps the useState hook to add logging. Hooks can be
+  /// composed -- meaning you can use hooks within hooks!
+  ValueNotifier<T> useLoggedState<T>([T initialData]) {
+    // First, call the useState hook. It will create a ValueNotifier for you that
+    // rebuilds the Widget whenever the value changes.
+    final result = useState<T>(initialData);
 
-  // Next, call the useValueChanged hook to print the state whenever it changes
-  useValueChanged<T, void>(result.value, (_, __) {
-    print(result.value);
-  });
+    // Next, call the useValueChanged hook to print the state whenever it changes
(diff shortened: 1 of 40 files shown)


noinskit:
Please consider this PR as a question/idea in the form of a working proof-of-concept.

As you might guess, this is a followup to a production bug when a widget class was imperfectly refactored from a HookWidget to a StatelessWidget with one instance of `useSomething` remaining. This was caught way too late in our case, and with this commit - it would've been a trivial compile-time error.

Note that it's not backward compatible in two ways:
1) Custom hooks need to be implemented this way as well.
2) HookBuilder's builder method has a different signature.

That said, incompatibility in a new major version might just be worth it if what it provides is more compile-time safety!

rrousselGit:
Thanks!

Honestly I'm not fan of the idea. It makes the syntax for creating functional hooks very heavy and removes the auto-complete. 
I don't like that `HookBuilder` and `HookWidget` behave differently either. 

At some point, hooks were exposed via a subclass of `BuildContext` for similar reasons. But people complained that the syntax was too heavy and started making an alternate implementation that uses static functions like now.

While I appreciate the effort, this would require an open issue to request feedback from the community, to see if they like this or not.

Personally, what I believe is the solution is instead a custom analyzer plugin – which implements the different hooks rules. That would fix _this_ specific issue, but also more such as "no condition" and co

noinskit:
It's a shame, but I understand.

I'm guessing that the previous, heavy syntax was before extension methods were there? I think they're what's making it acceptably non-invasive - it's actually just one line (plus another with just the closing brace) per hook definition.

I thought about the analyzer solution as well, but I leaned against it, because (I think) it no public API and also it doesn't feel right to me to use analyzer for something that the type system can handle by itself.

rrousselGit:
> I thought about the analyzer solution as well, but I leaned against it, because (I think) it no public API and also it doesn't feel right to me to use analyzer for something that the type system can handle by itself.


I believe that analyzer is correct, because hooks are in fact a language feature implemented using a workaround.

The only sad part is, custom analyzer plugins aren't very well integrated (they don't show up in `dart analyze`, you need a custom cli – but they do show up in the IDE fortunately )

rrousselGit:
As I said, I'd suggest making an issue first.

I'm not the only person to convince. If a good chunk of people like this, that's fine. I'm just saying that before, people were against it. But that was a long time ago. Opinions may have changed.

My main issue would be the difference between `HookWidget` and `HookBuilder` though, so I do lean to prefer subclassing `BuildContext` instead of extensions on `HookWidget` (and it'd work with [functional_widget](https://github.com/rrousselGit/functional_widget))

rrousselGit:
Closing as I don't think an extension on `HookWidget` is a good approach

PR #211, opened 2021-01-21
Null safety migration

DevNico:
General question since ValueNotifier has a non null generic i have implemented a second `useState` and `useValueNotifier` hook for nullable types: `useNullableState` and `useNullableValueNotifier`. Do you think this is the way to go or do you have a better idea? @rrousselGit

I also just removed all tests that are no longer necessary because of null safety.


DevNico:
The tests seem to be failing due to find.text working different in CI? Locally all tests pass.

rrousselGit:
I don't quite understand. What is the purpose of `useNullableState`? Why not do `useState<Model?>(null)`

DevNico:
It would be to make the initialState optional but I agree adding another hook would be unnecessary but the initialState wouldn't be optional anymore.

rrousselGit:
Indeed. I think removing them is better

rrousselGit:
Could you fix the CI too? Thanks!

DevNico:
I feel like the coverage diff is acceptable. LGTM @rrousselGit?

rrousselGit:
@DevNico could you give me the rights to make edits on your PR? Thanks!

lib/src/animation.dart
```diff
@@ -70,16 +70,16 @@ class _AnimationControllerHook extends Hook<AnimationController> {
...
-    this.upperBound,
-    this.vsync,
-    this.animationBehavior,
-    List<Object> keys,
+    required this.initialValue,
+    required this.lowerBound,
+    required this.upperBound,
+    required this.vsync,
+    required this.animationBehavior,
+    List<Object>? keys,
```
rrousselGit, line 78: Keys needs to be `List<Object?>? keys`

lib/src/async.dart
```diff
@@ -9,7 +9,7 @@ part of 'hooks.dart';
 ///   * [Future], the listened object.
 ///   * [useStream], similar to [useFuture] but for [Stream].
 AsyncSnapshot<T> useFuture<T>(Future<T> future,
-    {T initialData, bool preserveState = true}) {
+    {T? initialData, bool preserveState = true}) {
```
rrousselGit, line 12: initialData probably should be `required T initialData`

It's a breaking change, but matches with how `FutureProvider`/`StreamProvider` migrated

lib/src/async.dart
```diff
@@ -19,7 +19,7 @@ class _FutureHook<T> extends Hook<AsyncSnapshot<T>> {
 
   final Future<T> future;
```
rrousselGit, line 20: The Future (and Stream with useStream) should be nullable

lib/src/async.dart
```diff
@@ -29,14 +29,18 @@ class _FutureStateHook<T> extends HookState<AsyncSnapshot<T>, _FutureHook<T>> {
   /// An object that identifies the currently active callbacks. Used to avoid
   /// calling setState from stale callbacks, e.g. after disposal of this state,
   /// or after widget reconfiguration to a new Future.
-  Object _activeCallbackIdentity;
-  AsyncSnapshot<T> _snapshot;
+  Object? _activeCallbackIdentity;
+  late AsyncSnapshot<T> _snapshot;
```
rrousselGit, line 33: Prefer nullable properties instead of `late`.

That keyword is dangerous

lib/src/async.dart
```diff
@@ -158,38 +165,35 @@ class _StreamHookState<T> extends HookState<AsyncSnapshot<T>, _StreamHook<T>> {
   }
 
   void _subscribe() {
-    if (hook.stream != null) {
```
rrousselGit, line 161: those null checks shouldn't be removed

lib/src/async.dart
```diff
@@ -158,38 +165,35 @@ class _StreamHookState<T> extends HookState<AsyncSnapshot<T>, _StreamHook<T>> {
...
   }
 
   @override
   AsyncSnapshot<T> build(BuildContext context) {
     return _summary;
   }
 
-  AsyncSnapshot<T> initial() =>
-      AsyncSnapshot<T>.withData(ConnectionState.none, hook.initialData);
+  AsyncSnapshot<T> initial() => hook.initialData != null
```
rrousselGit, line 194: to remove after initialData is made required

lib/src/async.dart
DevNico: would you like to keep it as a named parameter or what do you think about either both `future` and `T initialData` as positioned parameter or both as named?

lib/src/async.dart
DevNico: > @DevNico could you give me the rights to make edits on your PR? Thanks!

The checkbox to allow edits from maintainers is already enabled or do you mean something else?

DevNico:
> @DevNico could you give me the rights to make edits on your PR? Thanks!

The checkbox to allow edits from maintainers is already enabled or do you mean something else?

lib/src/async.dart
DevNico: So would you want useFuture to return a nullable AsyncSnaphot (`AsyncSnapshot<T>? useFuture<T>`) or should the build method return a non nullable snapshot since it should be initialized anyway?

DevNico: How would you feel about
`late AsyncSnapshot<T> _summary = AsyncSnapshot<T>.withData(ConnectionState.none, hook.initialData);`?

rrousselGit: The return value must be non-nullable



> How would you feel about
> `late AsyncSnapshot<T> _summary = AsyncSnapshot<T>.withData(ConnectionState.none, hook.initialData);`?

That's fine, it's mostly uninitialized late variables that are scary

Then we can do the same thing with other hooks, like useStreamController

rrousselGit:
Weird, I've tried pushing some changes to your branch but the push was rejected

rrousselGit:
Thanks! I'll take over from here and will try to merge & publish it later today.

Do you happen to have a twitter handle to share? I'd want to thank you for your help 😄

DevNico:
Sure my handle is @DevNicoS even though I am not very active on Twitter 😅

rrousselGit:
Alright I'm going to release this. Thanks again for the help!

PR #221, opened 2021-03-07
Changing swapi.co to swapi.dev

subhendukundu:
https://swapi.co seems not accessible anymore. Switching to https://swapi.dev


rrousselGit:
Thanks!

Issue #225, opened 2021-03-18
Add postBuild event


jaumard:
**Is your feature request related to a problem? Please describe.**
I want to create a custom hook to replace `Observer` from mobx to his hook equivalent, from our discussion on slack you told me it was not possible until hooks have postBuild event. 

So I would like to work on this and make a PR, can you give me more details of what need to be done in order to add that event ? Some specs so I don't start wrong :) 

Thanks! Really love hooks!

jaumard:
I've seen this https://stackoverflow.com/questions/51216448/is-there-any-callback-to-tell-me-when-build-function-is-done-in-flutter 

So I'm wondering if the postBuild you referring is like this:
```
WidgetsBinding.instance.addPostFrameCallback((_) {
  hook.postBuild();
})
```

rrousselGit:
That won't work, because it is no longer "when the build method completed" but rather "when all widgets finished building"

jaumard:
Yeah I continued my search after finding that and created the PR with my new findings and it's indeed not using the  `addPostFrameCallback`

rrousselGit:
Closing as I don't plan on making such life-cycle after-all

PR #227, opened 2021-03-20
Add flutter_mobx_hooks and add didBuild event on hook state

jaumard:
fix https://github.com/rrousselGit/flutter_hooks/issues/225

Just a preview to see if I'm on the right track. 
Once we agree on the feature I'll do the associated tests


lib/src/framework.dart
```diff
@@ -417,6 +422,21 @@ mixin HookElement on ComponentElement {
       _buildCache = super.build();
     } finally {
       _isOptionalRebuild = null;
+
+      _hooks.map((entry) {
```
jaumard, line 426: Should I apply this to the full list or only to `_currentHookState ` quite don't understand why only the first one is used

lib/src/framework.dart
```diff
@@ -239,6 +239,10 @@ abstract class HookState<R, T extends Hook<R>> with Diagnosticable {
   @protected
   void didUpdateHook(T oldHook) {}
 
+  /// Called synchronously after the [HookWidget.build] method finished
+  @protected
+  void didBuild() {}
```
rrousselGit, line 244: I'm not fond of this API because 99.9% of hooks won't use this life-cycle

I'd like a way to do that, without having to iterate through **all** hooks as most of them don't care about this.

jaumard: What about having  a mixing `DidBuild` with that API ? We will be able to filter hooks that need postBuild and the once who don't.

jaumard: Another solution would be that the hook need to register to the didBuild event somehow. Let me know which one you prefer :) or if you have better ideas

rrousselGit: I'd prefer having something like "addListener" but for didBuild

jaumard: yeah that was my preferred solution, I'll look into this then :)

jaumard:
I've updated to have hooks registering to didBuild events, not pretty sure about the naming of stuff so you'll let me know your preferences.

I prefer `set` instead of `add` because add would mean you can have multiple callback which is not the case (and I don't see a need too)

lib/src/framework.dart
```diff
@@ -225,6 +226,12 @@ abstract class HookState<R, T extends Hook<R>> with Diagnosticable {
   @protected
   void initHook() {}
 
+  /// Register a callback for when [HookWidget.build] is finished
+  void setDidBuildListener(VoidCallback callback) {
+    _didBuildCallback = callback;
+    _element?.addHookForDidBuild(this);
```
rrousselGit, line 232: We'd need to override dispose to unsubscribe the hook

lib/src/framework.dart
```diff
@@ -417,6 +431,21 @@ mixin HookElement on ComponentElement {
       _buildCache = super.build();
     } finally {
       _isOptionalRebuild = null;
+
+      _needDidBuildHooks.map((hook) {
```
rrousselGit, line 435: for (final ... in )

rrousselGit:
That sounds alright.
Make sure to add tests with a legitimate use-case for it :smile:

In fact, it may make sense to add a `mobx_hooks` package.

jaumard:
> In fact, it may make sense to add a mobx_hooks package.

Do you want me to add it here ? I've started to make one based on the fork of mobX you have. But if you want it here I'll be happy to make another PR for it.

lib/src/framework.dart
jaumard: True, totally forget that we would need to remove them at some point, I'll add a `removeHookForDidBuild`

jaumard: To we want to call `removeHookForDidBuild ` automatically on the default `dispose` or do we let dev call it manually by overriding `dispose` ? For example if `_didBuildCallback` become nullable we can call `removeHookForDidBuild ` automatically on `dispose` if it's not null. It will prevent dev from forgetting to remove their hook in dispose

jaumard: I wanted to do something like this:
```
/// Equivalent of [State.dispose] for [HookState]
  @protected
  @mustCallSuper
  void dispose() {
    if (_didBuildCallback != null) {
      _didBuildCallback = null;
      _element?.removeHookForDidBuild(this);
    }
  }
```
But by doing so it triggers a lot of warning for existing Hook that doesn't call super. I prefer doing it automatically but if you think it's too much I'll just put on the doc that they should call remove on the on dispose.

jaumard:
> Make sure to add tests with a legitimate use-case for it 😄

That will be the trickiest as I don't see any other cases than mobx lool but to me what should be tested is that the didBuild event is well received.

davidmartos96:
> > Make sure to add tests with a legitimate use-case for it 😄
> 
> That will be the trickiest as I don't see any other cases than mobx lool but to me what should be tested is that the didBuild event is well received. 

I can see one use case not related with mobx. How about getting a RenderBox after the first build? Imagine you need the dimensions of the widget after it builds to trigger a different calculation based on that size.
Would this hook cover it?

lib/src/framework.dart
jaumard: I've not done that because if will be a change for all existing hooks, instead I have said on the doc that remove should be called if dispose is overridden

jaumard:
> I can see one use case not related with mobx. How about getting a RenderBox after the first build? Imagine you need the dimensions of the widget after it builds to trigger a different calculation based on that size.
Would this hook cover it?

I don't create any hook here, just give the possibility to hook to have a didBuild event. 

No it can't be used for this use case because dimensions are available only when all widgets have finished building, here the event is fired only for the current widget, no the all tree.

rrousselGit:
> I can see one use case not related with mobx. How about getting a RenderBox after the first build? Imagine you need the dimensions of the widget after it builds to trigger a different calculation based on that size.
> Would this hook cover it?

Arguably, this is a use-case for LayoutBuilder instead

And if we want to implement a `useLayout`, this wouldn't rely on a "postBuild" but instead about giving the ability for custom hooks to mount RenderObjects

davidmartos96:
@rrousselGit Yes, I was just wondering if that would be an alternate way to do it 😅 
I've actually created a custom hook that works fine for getting the widget size after build. It uses `addPostFrameCallback` under the hood.

```dart
Size useAfterBuildSize() {
  final size = useState<Size>(null);
  final context = useContext();
  useEffect(() {
    bool mounted = true;
    WidgetsBinding.instance.addPostFrameCallback(
      (timeStamp) {
        if (mounted) {
          final box = context.findRenderObject() as RenderBox;
          size.value = box.size;
        }
      },
    );
    return () => mounted = false;
  }, []);
  return size.value;
}

```

jaumard:
I'll work on flutter_mobx_hooks and push it here once is done :) let me some time ^^

jaumard:
@rrousselGit for having code coverage for flutter_mobx_hooks I think you need to set it up  right? because it stays pending...

rrousselGit:
It's probably just codecov that got stuck

packages/flutter_mobx_hooks/all_lint_rules.yaml
```diff
@@ -0,0 +1,181 @@
+linter:
+  rules:
```
rrousselGit, line 2: place it above `packages/` to be shared between all packages

packages/flutter_mobx_hooks/LICENSE
```diff
@@ -0,0 +1 @@
+TODO: Add your license here.
```
rrousselGit, line 1: Copy the license from flutter_hooks

packages/flutter_mobx_hooks/analysis_options.yaml
```diff
@@ -0,0 +1,75 @@
+include: all_lint_rules.yaml
+analyzer:
+  exclude:
+    - "**/*.g.dart"
+    - "**/*.freezed.dart"
+  strong-mode:
+    implicit-casts: false
```
rrousselGit, line 7: To be shared between packages too

rrousselGit:
Can you remove the android/ios/... folders from the example?

jaumard:
So look like codecov have shared the code coverage of both package for some reason...

packages/flutter_mobx_hooks/lib/src/use_autorun.dart
```diff
@@ -0,0 +1,91 @@
...
+    _run();
+  }
+
+  @override
+  void didUpdateHook(_AutorunHook oldHook) {
+    if (hook.context != oldHook.context) {
+      disposer();
+      _run();
+    }
+    // TODO(rrousselGit): hot reload name/delay/onError
```
jaumard, line 71: @rrousselGit any idea what you mean by this todo ? lol

packages/flutter_mobx_hooks/lib/src/use_observer.dart
```diff
@@ -0,0 +1,70 @@
...
+void useObserver({ReactiveContext? context}) {
+  final observer =
+      context == null ? const _ObserverHook() : _ObserverHook(context: context);
+  use(observer);
+}
+
+class _ObserverHook extends Hook<void> {
+  const _ObserverHook({ReactiveContext? context}) : _context = context;
+
+  // TODO(rrousselGit): scoped constructor
```
jaumard, line 18: same here :D

jaumard: if not I'll just remove those

packages/flutter_mobx_hooks/lib/src/use_reaction.dart
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  @override
+  void initHook() {
+    _createReaction();
+  }
+
+  @override
+  void didUpdateHook(_ReactionHook<T> oldHook) {
+    if (hook.context != oldHook.context) {
```
jaumard, line 79: no idea how I can test this @rrousselGit , any idea ? or use case because I don't really know why context could be except the main one

jaumard:
@rrousselGit if you don't mind I'll just remove the reactiveContext management, look like flutter_mobx always use the mainContext so no need to do more on our side too

packages/flutter_mobx_hooks/lib/src/use_autorun.dart
jaumard: code was removed

packages/flutter_mobx_hooks/lib/src/use_observer.dart
jaumard: removed

packages/flutter_mobx_hooks/lib/src/use_reaction.dart
jaumard: code remove as to me not relevant

jaumard:
ok all good all green finally :)

jaumard:
kind reminder @rrousselGit :)

rrousselGit:
I'll look at it next week-end. Busy working for now

rrousselGit:
Sorry for the delay, I'll look at it tonight

.github/workflows/build.yml
```diff
@@ -42,5 +43,5 @@ jobs:
         working-directory: packages/${{ matrix.package }}
 
       - name: Upload coverage to codecov
-        run: curl -s https://codecov.io/bash | bash
+        run: curl -s https://codecov.io/bash | bash -s -- -s coverage
```
rrousselGit, line 46: What is the purpose of this?

rrousselGit:
Could you fuse the analysis files in a single file at the root of the project?

packages/flutter_mobx_hooks/test/use_when_test.dart
```diff
@@ -0,0 +1,40 @@
...
+
+void main() {
+  testWidgets('useWhen', (tester) async {
+    var build = 0;
+    var value = 0;
+    final store = Counter();
+
+    await tester.pumpWidget(HookBuilder(
+      builder: (context) {
+        useWhen((_) => store.value == 2, () {
```
rrousselGit, line 16: Need to test what happens when the hook is updated with different callbacks

packages/flutter_mobx_hooks/test/use_reaction_test.dart
```diff
@@ -0,0 +1,33 @@
...
+
+void main() {
+  testWidgets('useReaction', (tester) async {
+    var build = 0;
+    var value = 0;
+    final store = Counter();
+
+    await tester.pumpWidget(HookBuilder(
+      builder: (context) {
+        useReaction<int>((_) => store.value, (newValue) {
```
rrousselGit, line 16: Need to test what happens when the hook is updated with different callbacks

rrousselGit: Also testing that on dispose the reaction is closed

packages/flutter_mobx_hooks/test/use_when_test.dart
rrousselGit: Also testing that on dispose the reaction is closed

packages/flutter_mobx_hooks/test/use_observer_test.dart
```diff
@@ -0,0 +1,26 @@
...
+import 'package:flutter_test/flutter_test.dart';
+
+import 'fake_store.dart';
+
+void main() {
+  testWidgets('useObserver', (tester) async {
+    var value = 0;
+    final store = Counter();
+
+    await tester.pumpWidget(HookBuilder(
```
rrousselGit, line 13: This needs to test more advanced scenarios, such as two nested HookBuilder, both using useObserver but observing different thing, and checking that rebuilding one doesn't rebuild the other

packages/flutter_mobx_hooks/test/use_autorun_test.dart
```diff
@@ -0,0 +1,33 @@
...
+
+void main() {
+  testWidgets('useAutorun', (tester) async {
+    var value = 0;
+    var build = 0;
+    final store = Counter();
+
+    await tester.pumpWidget(HookBuilder(
+      builder: (context) {
+        useAutorun((_) {
```
rrousselGit, line 16: Needs testing reaction disposal and callback updates

rrousselGit: Oh we also need to test that `_` parameter

.github/workflows/build.yml
jaumard: The purpose is to tell codecov to search for coverage directly in coverage folder, like this both package can provide their coverage and codecov is fusioning both

jaumard:
> Could you fuse the analysis files in a single file at the root of the project?

@rrousselGit you'r talking about analysis stuff ? It's already shared between the two project. Or I don't understand what you mean

jaumard:
@rrousselGit lots of test are failing for `hook_widget_test.dart` but none are failing locally... also stuff I didn't touch in that file  so weird it fails... do you have an idea ?

jaumard:
On my side it's all good, all tests for both packages pass, but can't find why on CI it fail for flutter_hook, any idea @rrousselGit ?

jaumard:
little bump here @rrousselGit :D

rrousselGit:
Sorry for the delay, I kind of forgot about this PR.

.github/workflows/build.yml
rrousselGit, line 46: The CI probably should start usinng `melos` like with Riverpod, to not have to have a `path:` dependency

packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -225,9 +226,26 @@ abstract class HookState<R, T extends Hook<R>> with Diagnosticable {
   @protected
   void initHook() {}
 
+  /// Register a callback for when [HookWidget] build is finished
+  /// If dispose method has been override, super.dispose or [removeDidBuildListener] must be called
+  void setDidBuildListener(VoidCallback callback) {
```
rrousselGit, line 231: This should likely be a getter/setter instead

Such that we do:

```dart
onDidBuild = () {

};

onDidBuild = null;
```

packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -353,6 +372,7 @@ mixin HookElement on ComponentElement {
 
   _Entry<HookState>? _currentHookState;
   final _hooks = LinkedList<_Entry<HookState>>();
+  final _needDidBuildHooks = <HookState>{};
```
rrousselGit, line 375: Could this be refactoring to a `List`, and have this property nullable by default? for performance

We wouldn't want to instantiate a list for every hook widget when few will actually care about this.

packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -417,6 +447,21 @@ mixin HookElement on ComponentElement {
       _buildCache = super.build();
     } finally {
       _isOptionalRebuild = null;
+
+      for (final hook in _needDidBuildHooks) {
```
rrousselGit, line 451: Will need to be wrapped into an `if (_needDidBuildHooks != null)` first

I'd also consider using `for (var i = 0; i < hook.length; i++)`, which should be slightly faster for a List.

rrousselGit:
Looking good, just a few small comments.

Do you think you could write a README?

jaumard:
Sure I'll do that asap :)

packages/flutter_hooks/lib/src/framework.dart
jaumard: Why a ´List` ? Why can't we keep the `Set` but make it nullable ?

packages/flutter_hooks/lib/src/framework.dart
jaumard: This is done

packages/flutter_hooks/lib/src/framework.dart
jaumard: This is done but I've kept the `Set`, let me know if you really want a list here.

packages/flutter_hooks/lib/src/framework.dart
jaumard: `if` is added but as I kept the the `Set` I've not change the `for`, let me know

packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -200,6 +200,7 @@ abstract class HookState<R, T extends Hook<R>> with Diagnosticable {
   @protected
   BuildContext get context => _element!;
   HookElement? _element;
+  VoidCallback? _onDidBuild;
```
rrousselGit, line 203: the private property should be next to the getter/setter for readability

packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -354,6 +373,7 @@ mixin HookElement on ComponentElement {
   _Entry<HookState>? _currentHookState;
   final _hooks = LinkedList<_Entry<HookState>>();
   final _shouldRebuildQueue = LinkedList<_Entry<bool Function()>>();
+  Set<HookState>? _needDidBuildHooks;
```
rrousselGit, line 376: Should be a list, not a set

The setter should be able to implement all the logic to avoid duplicate states.

jaumard: why doing that manually when the `Set` does it for us ? Because doing the manually on the setter is more code to test and maintain. What is the gain here ?

rrousselGit: Performance

List has a more efficient add & iteration

rrousselGit: Considering the deduplication logic doesn't require a `contains` call because the hook state knows if it was added to the element or not, using a List instead of a Set is a significant boost.

packages/flutter_hooks/lib/src/framework.dart
jaumard: Done

packages/flutter_hooks/lib/src/framework.dart
jaumard: Changed to List then :) thanks for the explanation!

packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -221,13 +221,32 @@ abstract class HookState<R, T extends Hook<R>> with Diagnosticable {
...
+
+  /// Callback for when [HookWidget] build is finished
+  VoidCallback? get onDidBuild => _onDidBuild;
+
+  /// Register a callback for when [HookWidget] build is finished
+  /// Pass null to unregister the callback
+  set onDidBuild(VoidCallback? callback) {
+    _onDidBuild = callback;
+    if (callback != null) {
+      _element?.addHookForDidBuild(this);
```
rrousselGit, line 234: We need to check that the previous callback was null too before making this call.

Otherwise `onDidBuild = x; onDidBUild = x;` will add `this` twice to the list

rrousselGit: I'd also do `_element!.`

And possibly start the setter with an `assert(_element != null, 'Cannot update onDidBuild inside the constructor of a HookState')`

jaumard: Done, I understand now the `List` instead of `Set` ^^

jaumard:
Readme is also done

.github/workflows/build.yml
jaumard: this is working, I would let's migrate to Melos in another PR please to avoir delaying this, I already don't know how to rebase master with the current state as everything moved...

jaumard:
Anything more needed on my side @rrousselGit ?

packages/flutter_hooks/lib/src/animation.dart
```diff
@@ -128,6 +128,7 @@ class _AnimationControllerHookState
 
   @override
   void dispose() {
+    super.dispose();
     _animationController.dispose();
```
rrousselGit, line 132:
```suggestion
    _animationController.dispose();
     super.dispose();
```

packages/flutter_hooks/lib/src/animation.dart
```diff
@@ -179,6 +180,7 @@ class _TickerProviderHookState
 
   @override
   void dispose() {
+    super.dispose();
```
rrousselGit, line 183: Needs to be at the end of the dispose method

packages/flutter_hooks/lib/src/async.dart
```diff
@@ -70,6 +70,7 @@ class _FutureStateHook<T> extends HookState<AsyncSnapshot<T?>, _FutureHook<T>> {
 
   @override
   void dispose() {
+    super.dispose();
     _unsubscribe();
```
rrousselGit, line 74:
```suggestion
    _unsubscribe();
    super.dispose();
```

packages/flutter_hooks/lib/src/async.dart
```diff
@@ -181,6 +182,7 @@ class _StreamHookState<T> extends HookState<AsyncSnapshot<T?>, _StreamHook<T>> {
...
```
rrousselGit, line 186:
```suggestion
    _unsubscribe();
    super.dispose();
```

packages/flutter_hooks/lib/src/async.dart
```diff
@@ -306,6 +308,7 @@ class _StreamControllerHookState<T>
 
   @override
   void dispose() {
+    super.dispose();
     _controller.close();
```
rrousselGit, line 312:
```suggestion
    _controller.close();
    super.dispose();
```

packages/flutter_hooks/lib/src/focus.dart
```diff
@@ -65,7 +65,10 @@ class _FocusNodeHookState extends HookState<FocusNode, _FocusNodeHook> {
   FocusNode build(BuildContext context) => _focusNode;
 
   @override
-  void dispose() => _focusNode.dispose();
+  void dispose() {
+    super.dispose();
+    _focusNode.dispose();
```
rrousselGit, line 70:
```suggestion
    _focusNode.dispose();
    super.dispose();
```

packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -221,13 +221,37 @@ abstract class HookState<R, T extends Hook<R>> with Diagnosticable {
...
+  VoidCallback? _onDidBuild;
+
+  /// Callback for when [HookWidget] build is finished
+  VoidCallback? get onDidBuild => _onDidBuild;
+
+  /// Register a callback for when [HookWidget] build is finished
+  /// Pass null to unregister the callback
+  set onDidBuild(VoidCallback? callback) {
+    assert(_element != null,
+        "Cannot update onDidBuild inside HookState's constructor");
```
rrousselGit, line 233:
```suggestion
    assert(
      _element != null,
      "Cannot update onDidBuild inside HookState's constructor",
    );
```

packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -417,6 +454,24 @@ mixin HookElement on ComponentElement {
       _buildCache = super.build();
     } finally {
       _isOptionalRebuild = null;
+
+      if (_needDidBuildHooks != null) {
+        for (var i = 0; i < _needDidBuildHooks!.length; i++) {
+          final hook = _needDidBuildHooks![i];
```
rrousselGit, line 460:
```suggestion
      final needDidBuildHooks = _needDidBuildHooks;
      if (needDidBuildHooks != null) {
        for (var i = 0; i < needDidBuildHooks.length; i++) {
          final hook = needDidBuildHooks[i];
```

packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -417,6 +454,24 @@ mixin HookElement on ComponentElement {
...
+          try {
+            hook.onDidBuild?.call();
+          } catch (exception, stack) {
+            FlutterError.reportError(FlutterErrorDetails(
+              exception: exception,
+              stack: stack,
+              library: 'hooks library',
+              context: ErrorDescription(
+                  'while calling `didBuild` on ${hook.runtimeType}'),
```
rrousselGit, line 469:
```suggestion
              context: ErrorDescription(
                'while calling `didBuild` on ${hook.runtimeType}',
              ),
```

packages/flutter_hooks/lib/src/listenable.dart
```diff
@@ -72,6 +72,7 @@ class _ListenableStateHook extends HookState<void, _ListenableHook> {
 
   @override
   void dispose() {
+    super.dispose();
     hook.listenable.removeListener(_listener);
```
rrousselGit, line 76:
```suggestion
    hook.listenable.removeListener(_listener);
    super.dispose();
```

packages/flutter_hooks/lib/src/page_controller.dart
```diff
@@ -49,7 +49,10 @@ class _PageControllerHookState
   PageController build(BuildContext context) => controller;
 
   @override
-  void dispose() => controller.dispose();
+  void dispose() {
+    super.dispose();
+    controller.dispose();
```
rrousselGit, line 54:
```suggestion
    controller.dispose();
    super.dispose();
```

packages/flutter_hooks/lib/src/primitives.dart
```diff
@@ -218,7 +218,10 @@ class _EffectHookState extends HookState<void, _EffectHook> {
   void build(BuildContext context) {}
 
   @override
-  void dispose() => disposer?.call();
+  void dispose() {
+    super.dispose();
+    disposer?.call();
```
rrousselGit, line 223:
```suggestion
    disposer?.call();
    super.dispose();
```

packages/flutter_hooks/lib/src/primitives.dart
```diff
@@ -278,6 +281,7 @@ class _StateHookState<T> extends HookState<ValueNotifier<T>, _StateHook<T>> {
 
   @override
   void dispose() {
+    super.dispose();
     _state.dispose();
```
rrousselGit, line 285:
```suggestion
    _state.dispose();
    super.dispose();
```

packages/flutter_hooks/lib/src/scroll_controller.dart
```diff
@@ -49,7 +49,10 @@ class _ScrollControllerHookState
   ScrollController build(BuildContext context) => controller;
 
   @override
-  void dispose() => controller.dispose();
+  void dispose() {
+    super.dispose();
+    controller.dispose();
```
rrousselGit, line 54:
```suggestion
    controller.dispose();
    super.dispose();
```

packages/flutter_hooks/lib/src/tab_controller.dart
```diff
@@ -51,7 +51,10 @@ class _TabControllerHookState
   TabController build(BuildContext context) => controller;
 
   @override
-  void dispose() => controller.dispose();
+  void dispose() {
+    super.dispose();
+    controller.dispose();
```
rrousselGit, line 56:
```suggestion
    controller.dispose();
    super.dispose();
```

packages/flutter_hooks/lib/src/text_controller.dart
```diff
@@ -87,7 +87,10 @@ class _TextEditingControllerHookState
   TextEditingController build(BuildContext context) => _controller;
 
   @override
-  void dispose() => _controller.dispose();
+  void dispose() {
+    super.dispose();
+    _controller.dispose();
```
rrousselGit, line 92:
```suggestion
    _controller.dispose();
    super.dispose();
```

packages/flutter_hooks/test/mock.dart
```diff
@@ -55,6 +55,7 @@ class HookStateTest<R> extends HookState<R?, HookTest<R>> {
 
   @override
   void dispose() {
+    super.dispose();
     hook.dispose?.call();
```
rrousselGit, line 59:
```suggestion
    hook.dispose?.call();
    super.dispose();
```

packages/flutter_mobx_hooks/coverage/html/src/use_autorun.dart.func.html
```diff
@@ -0,0 +1,72 @@
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
```
rrousselGit, line 1: Could you remove all folders besides `lib` and `test`, and add them to the gitignore?

jaumard: oops sure! I didn't notice they were committed sorry about that

jaumard: Should be all good now, don't know what append a lot of useless files where added in version source...

packages/flutter_hooks/lib/src/animation.dart
rrousselGit: This was resolved but still needs to be changed :)

jaumard: Oups i was thinking it was done by your suggested changes, I'll do that now

packages/flutter_mobx_hooks/lib/src/use_reaction.dart
```diff
@@ -0,0 +1,84 @@
...
+      equals: equals,
+      onError: onError,
+    ),
+  );
+}
+
+class _ReactionHook<T> extends Hook<void> {
+  const _ReactionHook(this.fn, this.effect,
+      {this.name, this.delay, this.fireImmediately, this.equals, this.onError});
+
```
rrousselGit, line 35:
```suggestion
  const _ReactionHook(
    this.fn,
    this.effect, {
    this.name,
    this.delay,
    this.fireImmediately,
    this.equals,
    this.onError,
  });

```

packages/flutter_hooks/lib/src/animation.dart
jaumard: Done now :)

jaumard:
Anything else needed @rrousselGit ? :)

jaumard:
Plop :D

jaumard:
@rrousselGit any way we can move forward with this ? ^^

jaumard:
ping @rrousselGit

rrousselGit:
Sorry for the super late reply.

I don't think I'll go with this after-all. I don't think modifying flutter_hooks for just this super specific use-case is reasonable, and I don't want to maintain the mobx hooks either.

If you still need it, I'd suggest publishing those hooks as a package on your own :)

PR #228, opened 2021-03-21
move to multi package structure

jaumard:


jaumard:
Yeah look like I didn't brake anything lol

rrousselGit:
packages/flutter_hooks/README appears to be broken

And the root README's logo appears to be broken too

.github/workflows/build.yml
```diff
@@ -13,6 +13,8 @@ jobs:
 
     strategy:
       matrix:
+        package:
+          - flutter_hooks
```
rrousselGit, line 17: leftover?

jaumard: No it's needed :) on next PR I'll add flutter_mobx_hooks in the matrix to have some CI

jaumard:
> packages/flutter_hooks/README appears to be broken
And the root README's logo appears to be broken too

It's broken because links to your repo, and as it's not merged files are 404

jaumard:
packages/flutter_hooks/README since broken indeed I'll check but logo it's because of path not being merged yet

jaumard:
My symlinks where wrong for some reasons... now it's ok! Good catch!

rrousselGit:
Looking good, thanks!

PR #239, opened 2021-04-23
Custom scroll controller

kranfix:
Adding support for scroll controllers that extends `ScrollController` like `FixedExtentScrollController`.

### DIFF
--- a/packages/flutter_hooks/lib/src/scroll_controller.dart
+++ b/packages/flutter_hooks/lib/src/scroll_controller.dart
@@ -1,49 +1,56 @@
 part of 'hooks.dart';
 
+class _ScrollControllerCreator {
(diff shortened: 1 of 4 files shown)


kranfix:
@rrousselGit, I commented the test `keys recreate hookstate` in the `hook_widget_test.dart` with a TODO requiring a fix to allow GitHub run actions.

rrousselGit:
I am not really fond of the idea, because that implies all other hooks should support "custom class"

I'd suggest making your own hook instead, which is fairly straightforward:

```dart
MyScrollController useMyScrollController() {
  final controller = useMemoized(() => MyScrollController());
  useEffect(() => controller.dispose, const []);
  useListenable(controller);
}
```

kranfix:
Ok. 😢

PR #241, opened 2021-05-04
Doc fixes

TNorbury:
Fixed some grammar and updated `useTextEditingController` doc comment to list the two setters that can be used to update the controller's value.


rrousselGit:
Awesome thanks!

Issue #229, opened 2021-03-21
useFuture/useStream: does initialData need to be required?


PiN73:
At 0.16.0 with nullsafety useFuture gets `required T initialData` parameter:
https://github.com/rrousselGit/flutter_hooks/blob/2b02b99ce6b327fd3a6df6f9c52cb2fe791600f6/lib/src/async.dart#L11-L15

This means when there is no initial data, one have to explicitly pass null:
```dart
final snapshot = useFuture(future, initialData: null);
```

Are there any benefits of making it required?

PiN73:
One possible benefit would be giving non-nullable type when initialData is provided:
```dart
Future<String> fetchData() {
  return Future.delayed(
    Duration(seconds: 2),
    () => 'Hello',
  );
}

class DisplayData extends HookWidget {
  @override
  Widget build(BuildContext context) {
    final AsyncSnapshot<String> snapshot = useFuture(
      useMemoized(fetchData),
      initialData: '...',
    );
    final String /*!*/ data = snapshot.data;
    return Text(data);
  }
}
```

But AsyncSnapshot class [defines](https://github.com/flutter/flutter/blob/8962f6dc68ec8e2206ac2fa874da4a453856c7d3/packages/flutter/lib/src/widgets/async.dart#L242) data as `T?` so we lose non-nullable type anyway

rrousselGit:
It is necessary, because otherwise `snapshot.requireData` would be `T?` instead of `T`

rrousselGit:
Considering the nature of hooks, you could use make a `useNullableFuture` that wraps `useFuture` to make `T` nullable

PiN73:
> It is necessary, because otherwise `snapshot.requireData` would be `T?` instead of `T`

Yes, but `T` doesn't seem stronger than `T` here because even if it's `T`, `snapshot.data` will be `T?` anyway

rrousselGit:
This isn't about "snapshot.data" but "snapshot.requireData"

PiN73:
Well, but my message is that `snapshot.data` is `T?` anyway. It is [defined](https://github.com/flutter/flutter/blob/8962f6dc68ec8e2206ac2fa874da4a453856c7d3/packages/flutter/lib/src/widgets/async.dart#L242) in flutter/lib/src/widgets/async.dart -> `class AsyncSnapshot<T>` as
```dart
final T? data;
```

kasperpeulen:
[...]

PR #243, opened 2021-05-18
add non-nullable variants for useFuture and useStream

PiN73:
If there is non-nullable `initialData`, this can be used to get `AsyncSnapshot` with non-nullable data. This will fix the problem described in https://github.com/rrousselGit/flutter_hooks/issues/229#issuecomment-843250650.

The whole hook implementation is reused. The only duplicating place is `use______` functions because return type can't depend on whether the optional argument is passed.

Also `return use(_FutureHook` in `useFuture` can be replaced with `return useFutureWithInitial` passing null to initialData. Then `_FutureHook` will be used exactly once, and variant will use hooks composition.

### DIFF
--- a/packages/flutter_hooks/lib/src/async.dart
+++ b/packages/flutter_hooks/lib/src/async.dart
@@ -22,7 +22,23 @@ AsyncSnapshot<T?> useFuture<T>(
   );
 }
 
-class _FutureHook<T> extends Hook<AsyncSnapshot<T?>> {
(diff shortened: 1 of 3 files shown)


rrousselGit:
I don't think that's a good workaround because of the added confusion of having two hooks for the same thing.

Aligning the implementation with StreamBuilder should be doable, so I'm going to close this. Thanks though!

PR #246, opened 2021-05-20
Change useStream Stream with nullable values

Butzlabben:
The Stream in `useStream()` should be able to accept null values, as there are streams (for example `FirebaseAuth.instance.authStateChanges()` returns `Stream<User?>`), which have null values in them. Also as the async snapshot already can return null, this shouldn't be a problem


rrousselGit:
Hello!

I will close this in favor of https://github.com/rrousselGit/flutter_hooks/pull/246 which matches what `StreamBuilder` does. The changes introduced by https://github.com/rrousselGit/flutter_hooks/pull/246 are incompatible with these changes.

PR #250, opened 2021-06-07
Fix async hooks nullability #229

### LINKED ISSUE #229: useFuture/useStream: does initialData need to be required?

davidmartos96:
Fixes #229 
As discussed, this now uses the strategy followed by the Flutter framework in `FutureBuilder` and `StreamBuilder`.
https://github.com/rrousselGit/flutter_hooks/issues/229#issuecomment-843835771

### DIFF
--- a/packages/flutter_hooks/lib/src/async.dart
+++ b/packages/flutter_hooks/lib/src/async.dart
@@ -8,7 +8,7 @@ part of 'hooks.dart';
 /// See also:
 ///   * [Future], the listened object.
 ///   * [useStream], similar to [useFuture] but for [Stream].
-AsyncSnapshot<T?> useFuture<T>(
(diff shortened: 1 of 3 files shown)


rrousselGit:
Looks good to me, thanks!

davidmartos96:
Thanks to you!

Issue #192, opened 2020-10-04
useAppLifecycleState Hook


yiss:
**Is your feature request related to a problem? Please describe.**
A hook to manage AppLifecycleState using WidgetBindingObserver. Documentation for WidgetBindingObserver : [WidgetBindingObserver](https://api.flutter.dev/flutter/widgets/WidgetsBindingObserver-class.htmll)

**Describe the solution you'd like**
I have few propositions to solve this :
#### The first :
Using an internal stream to control the value of the current AppLifecycle. the hook will return an `AsyncSnapshot<AppLifecycleState>`
```dart
AsyncSnapshot<AppLifecycleState> useAppLifecycleState();
...

class AppLifecycleStateExample extends HookWidget {
  @override
  Widget build(BuildContext context) {
    // Next, invoke the custom `useLoggedState` hook with a default value to
    // create a `counter` variable that contains a `value`. Whenever the value
    // is changed, this Widget will be rebuilt and the result will be logged!
    final appLifecycleState = useAppLifecycleState();
    return Scaffold(
      appBar: AppBar(
        title: const Text('AppLifecycleHook Example'),
      ),
      body: Center(
        // Read the current value from the counter
        child: Text('$appLifecycleState'),
      ),
    );
  }
}

```

**Describe alternatives you've considered**
#### 1st alternative : Expose the stream directly

```dart
Stream<AppLifecycleState> useAppLifecycleState();
```

#### 2nd alternative : Add an `onAppLifecycleChange` that will be triggered every time the AppLifecycleState changes.

```dart

typedef OnChangeLifecycleListener = void Function(AppLifecycleState state);

void useAppLifecycleState(OnChangeLifecycleListener onAppLifecycleChange);

```

#### 3rd alternative : same as the previous but instead a of just triggering the `onAppLifecycleChange` the method will also return the value ofAppLifecycleState :

```dart

typedef OnChangeLifecycleListener = AppLifecycleState Function(AppLifecycleState state);

AppLifecycleState useAppLifecycleState([OnChangeLifecycleListener onAppLifecycleChange]);

```

rrousselGit:
Sorry for the delay

You can make a custom class that implements the interface:

```dart
class MyObserver extends WidgetsBindingObserver {}
```

then use useMemoized/useEffect:

```dart
final myObserver = useMemoized(() => MyObserver());
useEffect(() {
  WidgetsBinding.instance.addObserver(myObserver);
[...]

PR #259, opened 2021-08-11
Added an `useAppStateLifecycle` hook

francipvb:
Hello,

I'm contributing this little hook to your project.

The justification is that sometimes we want the app lifecycle state for some use cases.

Recently I tried to use the camera plugin with hooks without success, so I'm contributing this hook to ease situations like this particular one.

Thank you for this library.


francipvb:
This fixes #192.

callmejm:
need this ! when can release ?

rrousselGit:
Thanks!

PR #276, opened 2021-12-13
feat: allow null in useListenable

### LINKED ISSUE #275: Why useListenable doesn't accept null?
**Is your feature request related to a problem? Please describe.**
```useListenable(this.propValue);```

When that `propValue` is `ValueNotifier<T>?`, I have trouble with using it with `useListenable` because `useListenable` doesn't allow nullable parameter. So I have a question about why it is not permitted.

**Describe the solution you'd like**
Allow null in `useListenable`.

**Describe alternatives you've considered**
There are some detours with the current signature.
-  Use`useValueNotifier()` for a placeholder listenable.
-  Separate as another hook widget and use that only if the listenable is not null.

But I think the above require tedious work in using that.

### PR DESCRIPTION
**Pull request by jeiea**
Close #275.

### DIFF
--- a/packages/flutter_hooks/lib/src/animation.dart
+++ b/packages/flutter_hooks/lib/src/animation.dart
@@ -24,7 +24,7 @@ class _UseAnimationStateHook extends _ListenableStateHook {
   String get debugLabel => 'useAnimation';
 
   @override
-  Object? get debugValue => (hook.listenable as Animation).value;
+  Object? get debugValue => (hook.listenable as Animation?)?.value;
 }
 
 /// Creates an [AnimationController] automatically disposed.
--- a/packages/flutter_hooks/lib/src/animation.dart
+++ b/packages/flutter_hooks/lib/src/animation.dart
@@ -24,7 +24,7 @@ class _UseAnimationStateHook extends _ListenableStateHook {
   String get debugLabel => 'useAnimation';
 
   @override
-  Object? get debugValue => (hook.listenable as Animation?)?.value;
+  Object? get debugValue => (hook.listenable as Animation).value;
 }
 
 /// Creates an [AnimationController] automatically disposed.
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: packages/flutter_hooks/test/use_listenable_test.dart [hunk 1]
```diff
@@ -72,6 +72,14 @@ void main() {
     expect(listenable.hasListeners, false);
 
     listenable.dispose();
+    listenable = null;
```
rrousselGit, line 75: Could you make a separate test instead of reusing an existing one? Tests should check only one thing at a time

jeiea: It makes sense. I pushed for that.

packages/flutter_hooks/lib/src/listenable.dart
```diff
@@ -24,7 +24,7 @@ class _UseValueListenableStateHook extends _ListenableStateHook {
   String get debugLabel => 'useValueListenable';
 
   @override
-  Object? get debugValue => (hook.listenable as ValueListenable).value;
+  Object? get debugValue => (hook.listenable as ValueListenable?)?.value;
```
rrousselGit, line 27:
```suggestion
  Object? get debugValue => (hook.listenable as ValueListenable).value;
```

packages/flutter_hooks/lib/src/animation.dart
```diff
@@ -24,7 +24,7 @@ class _UseAnimationStateHook extends _ListenableStateHook {
   String get debugLabel => 'useAnimation';
 
   @override
-  Object? get debugValue => (hook.listenable as Animation).value;
+  Object? get debugValue => (hook.listenable as Animation?)?.value;
```
rrousselGit, line 27:
```suggestion
  Object? get debugValue => (hook.listenable as Animation).value;
```

rrousselGit:
LGTM thanks!

PR #281, opened 2022-01-30
 Improve documentation

Renni771:
Hey there!  I took some time to fix some typos in the documentation and to make it more consistent with the style used in other parts of the codebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,7 +334,7 @@ They will take care of creating/updating/disposing an object.
 
 #### Misc hooks:
 
-These are a series of hooks with no single specific use-case.
+A series of hooks with no particular theme.
 
 | Name                                                                                                                                        | Description                                                                |
 | ------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------|
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 A Flutter implementation of React hooks: https://medium.com/@dan_abramov/making-sense-of-react-hooks-fdbde8803889
 
-Hooks are a new kind of object that manages a `Widget` life-cycles. They exist
+Hooks are a new kind of object that manage the life-cycle of a `Widget`. They exist
 for one reason: increase the code-sharing _between_ widgets by removing duplicates.
 
 ## Motivation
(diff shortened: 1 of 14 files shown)


packages/flutter_hooks/lib/src/async.dart
```diff
@@ -117,9 +117,9 @@ class _FutureStateHook<T> extends HookState<AsyncSnapshot<T>, _FutureHook<T>> {
   Object? get debugValue => _snapshot;
 }
 
-/// Subscribes to a [Stream] and return its current state in an [AsyncSnapshot].
+/// Subscribes to a [Stream] and returns its current state as an [AsyncSnapshot].
 ///
-/// * [preserveState] defines if the current value should be preserved when changing
+/// * [preserveState] determines if the current value should be preserved when changing
 /// the [Future] instance.
```
mhmdanas, line 123: I think this is supposed to be `Stream`?
```suggestion
/// the [Stream] instance.
```

packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -178,9 +178,9 @@ Calling them outside of build method leads to an unstable state and is therefore
     }
   }
 
-  /// Creates the mutable state for this hook linked to its widget creator.
+  /// Creates the mutable state for this [Hook] linked to its widget creator.
```
mhmdanas, line 181: According to https://dart.dev/guides/language/effective-dart/documentation#avoid-redundancy-with-the-surrounding-context, the class this method should be part of (`Hook` in this case) shouldn't be referenced here at all.

README.md
```diff
@@ -287,70 +287,68 @@ There are two ways to create a hook:
...
+| ----------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------|
+| [useListenable](https://pub.dartlang.org/documentation/flutter_hooks/latest/flutter_hooks/useListenable.html)           | Subscribes to a `Listenable` and marks the widget as needing build whenever the listener is called. |
+| [useValueNotifier](https://pub.dartlang.org/documentation/flutter_hooks/latest/flutter_hooks/useValueNotifier.html)     | Creates a `ValueNotifier` which will be automatically disposed.                                     |
+| [useValueListenable](https://pub.dartlang.org/documentation/flutter_hooks/latest/flutter_hooks/useValueListenable.html) | Subscribes to a `ValueListenable` and return its value.                                             |
 
-#### Misc
+#### Misc hooks:
 
-A series of hooks with no particular theme.
+These are a series of hooks with no single specific use-case.
```
rrousselGit, line 337:
```suggestion
A series of hooks with no particular theme.
```

The previous sentence sounds less confusing IMO. "use-case" has a different meaning than "theme"

packages/flutter_hooks/lib/src/framework.dart
rrousselGit: I don't personally agree with this warning. I prefer using [class] even when redundant, because we get proper lints if somehow the class was renamed

rrousselGit:
Thanks a lot!

Issue #287, opened 2022-03-08
Add rebuild method call to trigger build on HookWidget.


safield:
I have the usecase where I want a button to enable when a text field is not empty. To do this with flutter hooks, I have to create a superfluous state, using useState. When following the single source of truth principle, the only state I need to refer to is the textEditController.text.isNotEmpty to achieve the behaviour.

Observe the current implementation...

```
  Widget build(BuildContext context) {
    var ctrl = useTextEditingController();
    var superfluousCode = useState(0);
    ctrl.addListener(()=> superfluousCode.value++);
    VoidCallback? onPressed;
    if (ctrl.text.isNotEmpty) {
      onPressed = () => print("Pressed!");
    }
    return Scaffold(
      body: Column(
        children: [
          TextField(
            controller: ctrl,
          ),
          ElevatedButton(onPressed: onPressed, child: Text("Button")),
        ],
      )
    );
  }

```

In this example I need to create 'state' that is not needed, and would infact be code smell if I actually referred to the state instead of the actual source of truth which is 'ctrl.text.isNotEmpty'.

Ideally a HookWidget could just expose a method rebuild() to solve this issue.

ronnieeeeee:
Hi @safield ,
I don't know if this is the code you are expecting, but I wrote this code to enable the button widget when the text field is not empty. How do you like it?

```dart
class Example extends HookWidget {
  const Example({Key? key}) : super(key: key);


  @override
  Widget build(BuildContext context) {
    var ctrl = useTextEditingController();
    final textIsNotEmpty = useIsNotEmptyTextField(ctrl);
    return Scaffold(
        body: Column(
      children: [
        TextField(
          controller: ctrl,
        ),
        ElevatedButton(
            onPressed: textIsNotEmpty ? () => print("Pressed!") : null,
            child: Text("Button")),
      ],
    ));
  }
}


bool useIsNotEmptyTextField(TextEditingController textEditingController) {
  final hasNotEmpty = useState(textEditingController.text.isNotEmpty);


  useEffect(() {
    void l() {
      hasNotEmpty.value = textEditingController.text.isNotEmpty;
    }


    textEditingController.addListener(l);
    return () => textEditingController.removeListener(l);
  }, [textEditingController]);


  return hasNotEmpty.value;
}
```

For more information, please try this repository in run "main.dart" .
https://github.com/ronnieeeeee/flutter_hooks287

safield:
[...]

PR #292, opened 2022-04-03
Add useListenableMap

ronnieeeeee:
Hi @rrousselGit 😄
I developed useListenableMap() [suggested](https://github.com/rrousselGit/flutter_hooks/issues/287#issuecomment-1061818631) by Remi.

```dart
class Example extends HookWidget {
  @override
   final listenable = useTextEditingController();
   final bool textIsEmpty =
        useListenableMap(listenable, () => listenable.text.isEmpty);
   return Column(
      children: [
        TextField(controller: listenable),
        ElevatedButton(
            // if textIsEmpty is false,　the button is disabled.
            onPressed: textIsEmpty ? null : () => print("Pressed!"),
            child: Text("Button")),
      ],
    );
  }
}
```

### About useListenableMap()
Subscribe to a Listenable, when [Listenable] update execute callback(). the result is the return value.
It doesn't rebuild the widget unless that return of callback() result does not change. 
An unnecessary rebuild does not occur so do not unnecessary performance.

There was a[ problem](https://github.com/rrousselGit/flutter_hooks/issues/287) that writing the same process without useListenableMap would be verbose, so I created it.

### DIFF
--- a/packages/flutter_hooks/lib/src/listenable_map.dart
+++ b/packages/flutter_hooks/lib/src/listenable_map.dart
@@ -1,0 +1,70 @@
+part of 'hooks.dart';
+
+/// Subscribes to a [Listenable]
+/// and execute callback() whenever [Listenable] update.
+/// Subscribe to a Listenable.
+/// When [Listenable] update execute callback(). it result value is return value.
+/// It don't rebuild the widget unless that result of callback() does not changed. so an unnecessary rebuild not occur, performance is better.
+///
+/// ```dart
+/// class Example extends HookWidget {
+///   @override
+///   Widget build(BuildContext context) {
+///    final listenable = useTextEditingController();
+///    final bool textIsEmpty =
+///         useListenableMap(listenable, () => listenable.text.isEmpty);
+///    return Column(
+///       children: [
+///         TextField(controller: listenable),
+///         ElevatedButton(
+///             // if textIsEmpty is false,　the button is disabled.
+///             onPressed: textIsEmpty ? null : () => print("Pressed!"),
+///             child: Text("Button")),
+///       ],
+///     );
+///   }
+/// }
+/// ```
+R useListenableMap<T extends Listenable, U extends R Function(), R>(
+    T listenable, U callback) {
+  return use(_ListenableMapHook<T, U, R>(listenable, callback)).value;
+}
+
+class _ListenableMapHook<T extends Listenable, U extends R Function(), R>
+    extends Hook<ValueNotifier<R>> {
+  const _ListenableMapHook(this.listenable, this.callback);
+
+  final T listenable;
+  final U callback;
+
+  @override
+  _ListenableMapHookState<T, U, R> createState() =>
+      _ListenableMapHookState<T, U, R>();
+}
+
+class _ListenableMapHookState<T extends Listenable, U extends R Function(), R>
+    extends HookState<ValueNotifier<R>, _ListenableMapHook<T, U, R>> {
+  late final ValueNotifier<R> _state = ValueNotifier<R>(hook.callback())
+    ..addListener(() => setState(() {}));
+
+  @override
+  void initHook() {
+    super.initHook();
+    hook.listenable.addListener(_listener);
+  }
+
+  @override
+  ValueNotifier<R> build(BuildContext context) => _state;
+
+  void _listener() {
+    _state.value = hook.callback();
+  }
+
+  @override
+  void dispose() {
+    hook.listenable.removeListener(_listener);
+  }
+
+  @override
+  String get debugLabel => 'useListenableMap<$T>';
+}
--- a/packages/flutter_hooks/lib/src/listenable_map.dart
+++ b/packages/flutter_hooks/lib/src/listenable_map.dart
@@ -25,25 +25,22 @@ part of 'hooks.dart';
 ///   }
 /// }
 /// ```
-R useListenableMap<T extends Listenable, U extends R Function(), R>(
-    T listenable, U callback) {
-  return use(_ListenableMapHook<T, U, R>(listenable, callback)).value;
+R useListenableMap<R>(Listenable listenable, R Function() callback) {
+  return use(_ListenableMapHook<R>(listenable, callback)).value;
 }
 
-class _ListenableMapHook<T extends Listenable, U extends R Function(), R>
-    extends Hook<ValueNotifier<R>> {
+class _ListenableMapHook<R> extends Hook<ValueNotifier<R>> {
   const _ListenableMapHook(this.listenable, this.callback);
 
-  final T listenable;
-  final U callback;
+  final Listenable listenable;
+  final R Function() callback;
 
   @override
-  _ListenableMapHookState<T, U, R> createState() =>
-      _ListenableMapHookState<T, U, R>();
+  _ListenableMapHookState<R> createState() => _ListenableMapHookState<R>();
 }
 
-class _ListenableMapHookState<T extends Listenable, U extends R Function(), R>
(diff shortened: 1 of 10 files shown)


rrousselGit:
Thanks, awesome work 😄 

I don't like the name though. It's not really a "map".
Maybe useListenableSelector?

packages/flutter_hooks/lib/src/listenable_map.dart
```diff
@@ -0,0 +1,70 @@
...
+///         ElevatedButton(
+///             // if textIsEmpty is false,　the button is disabled.
+///             onPressed: textIsEmpty ? null : () => print("Pressed!"),
+///             child: Text("Button")),
+///       ],
+///     );
+///   }
+/// }
+/// ```
+R useListenableMap<T extends Listenable, U extends R Function(), R>(
```
rrousselGit, line 28: That's too many generics.

Only R should be necessary, the others shouldn't be here.

packages/flutter_hooks/lib/src/listenable_map.dart
```diff
@@ -0,0 +1,70 @@
...
+  final U callback;
+
+  @override
+  _ListenableMapHookState<T, U, R> createState() =>
+      _ListenableMapHookState<T, U, R>();
+}
+
+class _ListenableMapHookState<T extends Listenable, U extends R Function(), R>
+    extends HookState<ValueNotifier<R>, _ListenableMapHook<T, U, R>> {
+  late final ValueNotifier<R> _state = ValueNotifier<R>(hook.callback())
```
rrousselGit, line 47: Don't use a ValueNotifier. Use a property and call setState on change

packages/flutter_hooks/test/use_listenable_map_test.dart
```diff
@@ -0,0 +1,81 @@
...
+
+    listenable.value++;
+    await tester.pump();
+    expect(listenable.value, 46);
+    expect(isOdd, false);
+    expect(element.dirty, false);
+
+    await tester.pumpWidget(const SizedBox());
+    // ignore: invalid_use_of_protected_member
+    expect(() => listenable.hasListeners, throwsFlutterError);
```
rrousselGit, line 79: Why should this line be throwing?

The hook listenable isn't disposed, so calling hasListeners shouldn't throw

packages/flutter_hooks/test/use_listenable_map_test.dart
```diff
@@ -0,0 +1,81 @@
...
+void main() {
+  testWidgets('debugFillProperties', (tester) async {
+    await tester.pumpWidget(HookBuilder(
+      builder: (context) {
+        final listenable = ValueNotifier<int>(42);
+        useListenableMap<ValueNotifier<int>, bool Function(), bool>(
+            listenable, () => listenable.value.isOdd);
+        return const SizedBox();
+      },
+    ));
```
rrousselGit, line 17: Nit: add trailing comma and format

packages/flutter_hooks/test/use_listenable_map_test.dart
```diff
@@ -0,0 +1,81 @@
...
+import 'package:flutter_hooks/src/hooks.dart';
+import 'mock.dart';
+
+void main() {
+  testWidgets('debugFillProperties', (tester) async {
+    await tester.pumpWidget(HookBuilder(
+      builder: (context) {
+        final listenable = ValueNotifier<int>(42);
+        useListenableMap<ValueNotifier<int>, bool Function(), bool>(
+            listenable, () => listenable.value.isOdd);
```
rrousselGit, line 14: Add trailing comma

packages/flutter_hooks/test/use_listenable_map_test.dart
```diff
@@ -0,0 +1,81 @@
...
+      equalsIgnoringHashCodes(
+        'HookBuilder\n'
+        ' │ useListenableMap<ValueNotifier<int>>:\n'
+        ' │   ValueNotifier<bool>#00000(false)\n'
+        ' └SizedBox(renderObject: RenderConstrainedBox#00000)\n'
+        '',
+      ),
+    );
+  });
+  testWidgets('useListenableMap', (tester) async {
```
rrousselGit, line 34: Add empty line before test

packages/flutter_hooks/lib/src/listenable_map.dart
ronnieeeeee: > Thanks, awesome work 😄
> 
> I don't like the name though. It's not really a "map". Maybe useListenableSelector?

Thank you, Remi 😄
I'm so glad to hear you say that!
I prefer useListenableSelector as naming. I am rethinking the naming.

Thank you for your many suggestions for my code, even though you seem to be busy. I will fix it.

packages/flutter_hooks/test/use_listenable_map_test.dart
ronnieeeeee: I have too much knowledge to learn, so I will study. Thank you, Remi 👍

ronnieeeeee:
> 

I couldn't think of good naming, so to name useListenableSelector.

packages/flutter_hooks/lib/src/listenable_selector.dart
```diff
@@ -0,0 +1,71 @@
...
+    hook.listenable.addListener(_listener);
+  }
+
+  @override
+  R build(BuildContext context) => _state;
+
+  void _listener() {
+    final result = hook.callback();
+    if (_state != result) {
+      _state = result;
```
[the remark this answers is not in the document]
ronnieeeeee: How embarrassing!
I was referring only to primitive.dart and listenable.dart.
So without thinking, I assumed setState would be used like this `setState(() {});`.

rrousselGit:
Oh, I spotted a few problems. 🙈 
I'll revert the PR. It'd be great if you can have a look @ronnieeeeee 

The problems being:
- we need to handle `listenable` changing
- similarly, when the widget rebuilds, we need to re-invoke the callback to update the `_state`
  (because users could write `useListenableSelector(listenable, () => listenable.value + widget.param)`

ronnieeeeee:
Hi Remi(@rrousselGit), I am sorry that the implementation of the useListeanbleSelector did not work,
even though you took time out of your busy responding to other issues and PRs.
And thank you for all the fixes and suggestions.
However, I am very interested in flutter_hooks (and riverpod) and enjoy being involved in its development, 
so I will continue to implement useListeanbleSelector. 👍

ronnieeeeee:
I am thinking of adding to useListenableSelector the functionality that useEffect and others implement by taking keys as the second argument.

The plan looks like this.
```dart
useListenableSelector(listenable, () => listenable.value + widget.param, [widget.param])
useListenableSelector(listenable, () => listenable.value + widget.param, [listenable,widget.param])
```

ronnieeeeee:
Or consider also useListenableReducer.

rrousselGit:
> I am thinking of adding to useListenableSelector the functionality that useEffect and others implement by taking keys as the second argument.
> 
> The plan looks like this.
> 
> ```dart
> useListenableSelector(listenable, () => listenable.value + widget.param, [widget.param])
> useListenableSelector(listenable, () => listenable.value + widget.param, [listenable,widget.param])
> ```

I don't think this make sense for this hook.

I'd expect to simply override `didUpdateHook` and handle changes

rrousselGit:
Selectors typically don't act as a cache. It's fine to call the callback again when the hook rebuilds

If folks want a cache, they should use useMemoized

ronnieeeeee:
> 

That makes sense、I'll try it.

ronnieeeeee:
> Selectors typically don't act as a cache. It's fine to call the callback again when the hook rebuilds

I surely agree.
Thank you for the advice.

Issue #101, opened 2019-09-11
AutomaticKeepAliveClientMixin for HookWidget?


zaralockheart:
Hi, Thanks for the great library. Can I know if there's any way I can use `AutomaticKeepAliveClientMixin` for widget that extends `HookWidget`?

rrousselGit:
No. On the other hand, you can copy its implementation and make it a `useAutomaticKeepAlive` hook.

zaralockheart:
:( I really hope there is plan for this. Anyway thanks! :D

xantiagoma:
@rrousselGit  Any update about this? or how should be the implementation for `useAutomaticKeepAlive`?

DevNico:
I wrote a hook that seems to work as far as I tested.
```dart
import 'package:flutter/widgets.dart';
import 'package:flutter_hooks/flutter_hooks.dart';

void useAutomaticKeepAlive({
  bool wantKeepAlive,
}) =>
    use(_AutomaticKeepAliveHook(
      wantKeepAlive: wantKeepAlive ?? true,
    ));

class _AutomaticKeepAliveHook extends Hook<void> {
  final bool wantKeepAlive;

  _AutomaticKeepAliveHook({this.wantKeepAlive});

  @override
  HookState<void, _AutomaticKeepAliveHook> createState() => _AutomaticKeepAliveHookState();
}

class _AutomaticKeepAliveHookState extends HookState<void, _AutomaticKeepAliveHook> {
  KeepAliveHandle _keepAliveHandle;

  void _ensureKeepAlive() {
    assert(_keepAliveHandle == null);
    _keepAliveHandle = KeepAliveHandle();
    KeepAliveNotification(_keepAliveHandle).dispatch(context);
  }

  void _releaseKeepAlive() {
    _keepAliveHandle.release();
    _keepAliveHandle = null;
  }

  void updateKeepAlive() {
    if (hook.wantKeepAlive) {
      if (_keepAliveHandle == null) _ensureKeepAlive();
    } else {
      if (_keepAliveHandle != null) _releaseKeepAlive();
    }
  }

  @override
  void initHook() {
    super.initHook();
    if (hook.wantKeepAlive) _ensureKeepAlive();
  }

  @override
  void build(BuildContext context) {
    if (hook.wantKeepAlive && _keepAliveHandle == null) _ensureKeepAlive();
    return null;
  }

  @override
  void deactivate() {
[...]

PR #297, opened 2022-04-11
add useAutomaticKeepAlive hook

DevNico:
See #101


DevNico:
@rrousselGit What do you think?

rrousselGit:
I'll review this later

packages/flutter_hooks/test/use_automatic_keep_alive_test.dart
```diff
@@ -0,0 +1,131 @@
...
+            textDirection: TextDirection.ltr,
+            child: DefaultTabController(
+              length: 2,
+              child: TabBarView(
+                children: [
+                  HookBuilder(builder: (context) {
+                    final wantKeepAlive = useState(false);
+                    useAutomaticKeepAlive(wantKeepAlive: wantKeepAlive.value);
+                    useEffect(() {
+                      Future.delayed(
```
rrousselGit, line 84: Don't rely on delays inside tests. 

Could you refactor the test to not do that? :pray:

DevNico: The delay is matched to the pump duration but I'll try to change that.

rrousselGit: My suggestion would be to have a ValueNotifier in the test (outside of the widget)

and do useValueListenable(valueNotifier)

Then change the pump into a valueNotifier.value = true;

DevNico: Great suggestion, works like a charm.

DevNico:
Should I remove the `!` since your ci seems to be on a newer flutter version?

rrousselGit:
This would require bumping the minimum Flutter SDK.

Maybe add an ignore for now, until the fix can be applied to stable.

DevNico:
friendly ping @rrousselGit

kNoAPP:
🙏 @rrousselGit

PR #348, opened 2023-03-15
ko_KR translate

BGM-109:
The currently uploaded kr document is just Google Translator.
It's very uncomfortable to read.
I tried to translate it simply as I read it.
I'm not a professional translator, but I think it'll be more helpful than now.
Thank you for making the hook. I'm using it well.


rrousselGit:
Thanks!

PR #369, opened 2023-07-10
Make useCallback hook's keys optional

jezsung:
Makes the `useCallback` hook's `keys` property optional and default to an empty list.

### DIFF
--- a/packages/flutter_hooks/test/use_callback_test.dart
+++ b/packages/flutter_hooks/test/use_callback_test.dart
@@ -4,6 +4,42 @@ import 'package:flutter_hooks/flutter_hooks.dart';
 import 'mock.dart';
 
 void main() {
+  testWidgets('useCallback', (tester) async {
+    late int Function() fn;
+
+    await tester.pumpWidget(
+      HookBuilder(builder: (context) {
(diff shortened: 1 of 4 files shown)


rrousselGit:
Thanks!
Could you add a test for this too? :pray:

jezsung:
@rrousselGit Sure, should I also upgrade the version to `0.18.7`?

rrousselGit:
No, I bump versions myself

packages/flutter_hooks/test/use_callback_test.dart
```diff
@@ -0,0 +1,36 @@
+import 'package:flutter/widgets.dart';
+import 'package:flutter_hooks/flutter_hooks.dart';
+
+import 'mock.dart';
+
+void main() {
+  testWidgets(
```
rrousselGit, line 7: I like that we make a new file for this. But could you move existing useCallback tests in here too then?
https://github.com/rrousselGit/flutter_hooks/blob/2a6b45cc5c893e14da69cab7485047c7f1e30f2f/packages/flutter_hooks/test/memoized_test.dart#L68

rrousselGit:
LGTM, thanks!

PR #373, opened 2023-07-15
Add useStreamSubscription hook

jezsung:
Adds a new hook `useStreamSubscription` that allows you to subscribe to a Stream and register callback handlers such as `onData`, `onError`, and `onDone`.

Returns the `StreamSubscription` returned by the `Stream.listen`.

**CAUTION**
Awaiting on the returned `StreamSubscription.cancel` never ends on a test. Not sure if this is somehow expected.

https://github.com/jezsung/flutter_hooks/blob/53ed52bf3605952f2eb790d8c57b2831684b2479/packages/flutter_hooks/test/use_stream_subscription_test.dart#L207-L209

### DIFF
--- a/packages/flutter_hooks/lib/src/async.dart
+++ b/packages/flutter_hooks/lib/src/async.dart
@@ -392,8 +392,9 @@ class _StreamListenerHookState<T>
   }
 
   void _subscribe() {
-    if (hook.stream != null) {
-      _subscription = hook.stream!.listen(
+    final stream = hook.stream;
+    if (stream != null) {
+      _subscription = stream.listen(
         hook.onData?.call,
         onError: hook.onError?.call,
         onDone: hook.onDone?.call,
--- a/packages/flutter_hooks/lib/src/async.dart
+++ b/packages/flutter_hooks/lib/src/async.dart
@@ -395,9 +395,9 @@ class _StreamListenerHookState<T>
     final stream = hook.stream;
     if (stream != null) {
       _subscription = stream.listen(
-        hook.onData?.call,
-        onError: hook.onError?.call,
-        onDone: hook.onDone?.call,
+        hook.onData,
+        onError: hook.onError,
+        onDone: hook.onDone,
         cancelOnError: hook.cancelOnError,
       );
     }
--- a/packages/flutter_hooks/lib/src/async.dart
+++ b/packages/flutter_hooks/lib/src/async.dart
@@ -320,3 +320,95 @@ class _StreamControllerHookState<T>
   @override
   String get debugLabel => 'useStreamController';
 }
+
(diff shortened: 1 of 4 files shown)


rrousselGit:
Hello!
What's the difference with the useStreamListener PR?

I don't like the naming on this hook.
Generally such hooks would be named `useOnStreamChange`

packages/flutter_hooks/test/use_stream_subscription_test.dart
```diff
@@ -0,0 +1,220 @@
...
+        element
+            .toDiagnosticsNode(style: DiagnosticsTreeStyle.offstage)
+            .toStringDeep(),
+        equalsIgnoringHashCodes(
+          'HookBuilder\n'
+          " │ useStreamSubscription: Instance of '_ControllerSubscription<int>'\n"
+          ' └SizedBox(renderObject: RenderConstrainedBox#00000)\n',
+        ),
+      );
+    },
```
rrousselGit, line 36: Don't add a trailing comma for tests

rrousselGit:
If the goal is to return a StreamSubscription to cancel the subscription early, that's fine. Although other hooks don't really do this.



Could you add this to the README?

rrousselGit:
Could you add 100% coverage too?

jezsung:
Just another thought on this, how about using a callable class to allow syntax like this:
```dart
final StreamSubscription subscription = useStream.listen(
  onData: (data) => {},
);
```
I saw the `useTextEditingController` is using a callable class to make this syntax possible and it seems really cool.

packages/flutter_hooks/lib/src/async.dart
```diff
@@ -320,3 +320,97 @@ class _StreamControllerHookState<T>
...
+  void dispose() {
+    _unsubscribe();
+  }
+
+  void _subscribe() {
+    if (hook.stream != null) {
+      _subscription = hook.stream!.listen(
+        hook.onData?.call,
+        onError: hook.onError?.call,
+        onDone: hook.onDone?.call,
```
rrousselGit, line 399:
```suggestion
        hook.onData,
        onError: hook.onError,
        onDone: hook.onDone,
```

packages/flutter_hooks/lib/src/async.dart
```diff
@@ -320,3 +320,97 @@ class _StreamControllerHookState<T>
...
+  }
+
+  @override
+  void dispose() {
+    _unsubscribe();
+  }
+
+  void _subscribe() {
+    if (hook.stream != null) {
+      _subscription = hook.stream!.listen(
```
rrousselGit, line 396:
```suggestion
    final stream = hook.stream;
    if (stream != null) {
      _subscription = stream.listen(
```

rrousselGit:
LGTM, thanks!

Issue #155, opened 2020-07-13
use DeepCollectionEquality for hook keys


rrousselGit:

jezsung:
@rrousselGit 
I was looking for how equality works for hook keys and found this issue. Not quite sure if using deep collection equality by default is a good idea. Is this how React Hook works?

rrousselGit:
I don't remember. And honestly at that point, this issue doesn't really matter.

I'm not sure it's worth making a change now.

jezsung:
React Hook uses [`Object.is()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) comparison which is not the same as the `DeepCollectionEquality`.

I think this change is unnecessary.

jezsung:
The only differences between `Object.is()` and Dart `==` comparison are the following:

- `NaN` comparison
- `0` and `-0` comparison (Dart doesn't have `+0`)

`Object.is()`
```js
Object.is(NaN, NaN) // true
Object.is(0, -0) // false
```

Dart `==`
```dart
double.nan == double.nan // false
0 == -0 // true
```

In order to exactly mimic the behavior of React Hook, the following changes seem to be needed:

``` dart
// Special case for NaN
if (a.isNaN && b.isNaN) {
  return true;
}

// Special case for 0 and -0
if (a == 0 && b == 0) {
  return (1 / a) == (1 / b);
}
```

Technically, this would be a breaking change but since comparing those values are quite rare, could be worth it to make this change. What do you think @rrousselGit?

rrousselGit:
These sound like valuable changes.

For 0/-0, I'd rely on .isNegative

jezsung:
`isNegative` seems better. I'll make a PR shortly

PR #374, opened 2023-07-19
Mimic React Hook's dependencies comparison

jezsung:
[#155 ](https://github.com/rrousselGit/flutter_hooks/issues/155#issuecomment-1641688259)

This PR changes the comparison behavior to mimic the React Hook dependencies comparison which uses [`Object.is()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is).

Previously, `keys` are compared by the `==` operator. The Dart `==` operator works differently from the JavaScript `Object.is()`:

Object.is()
```js
Object.is(NaN, NaN) // true
Object.is(0, -0) // false
```
Dart ==
```dart
double.nan == double.nan // false
0 == -0 // true
```

This PR checks if a list of `keys` contain a value with the type `num` and proceeds to handle cases where the value is either `NaN` or `0.0` and `-0.0`.

### DIFF
--- a/packages/flutter_hooks/lib/src/framework.dart
+++ b/packages/flutter_hooks/lib/src/framework.dart
@@ -172,7 +172,23 @@ Calling them outside of build method leads to an unstable state and is therefore
       if (!i1.moveNext() || !i2.moveNext()) {
         return true;
       }
-      if (i1.current != i2.current) {
+
+      final curr1 = i1.current;
+      final curr2 = i2.current;
+
+      if (curr1 is num && curr2 is num) {
+        // Checks if both are NaN
+        if (curr1.isNaN && curr2.isNaN) {
+          return true;
+        }
+
+        // Checks if one is 0.0 and the other is -0.0
+        if (curr1 == 0 && curr2 == 0) {
+          return curr1.isNegative == curr2.isNegative;
+        }
+      }
+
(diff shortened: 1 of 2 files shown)


packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -172,7 +172,23 @@ Calling them outside of build method leads to an unstable state and is therefore
       if (!i1.moveNext() || !i2.moveNext()) {
         return true;
       }
-      if (i1.current != i2.current) {
+
```
rrousselGit, line 175: The docs need to be updated to match

rrousselGit:
LGTM, with a tiny bit of documentation needed

Could you also add a changelog?

jezsung:
Documenting has been done! You might wanna resolve the `CHANGELOG.md` conflict. I'll just leave it for you.

packages/flutter_hooks/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## Unreleased minor
```
rrousselGit, line 1: It's a breaking change
```suggestion
## Unreleased major
```

rrousselGit:
Looking good, thanks!

PR #377, opened 2023-08-15
Feature: ability to use nullable listener in useListenableSelector

JWo1F:
## Description

This Pull Request adds the ability to pass `null` as a parameter to the `useListenableSelector` hook. Previously, you could only pass a `Listener` instance, but now you can also pass `Listener?` to the function.

This change is necessary when the listener is initialized later, for example, on the second render. By allowing `null` as a parameter, developers can delay the initialization of the listener until it's ready.

## Example Usage

```dart
// Before
useListenableSelector<SomeListener>(myListener, () => myListener.data);

// After
useListenableSelector<SomeListener?>(null, () => myListener?.data);

### DIFF
--- a/packages/flutter_hooks/lib/src/listenable_selector.dart
+++ b/packages/flutter_hooks/lib/src/listenable_selector.dart
@@ -25,7 +25,7 @@ part of 'hooks.dart';
 ///
 
 R useListenableSelector<R>(
-  Listenable listenable,
+  Listenable? listenable,
   R Function() selector,
 ) {
   return use(_ListenableSelectorHook(listenable, selector));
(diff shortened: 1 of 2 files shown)


rrousselGit:
Nice. Could you add tests for this?

rrousselGit (changes requested):
Needs testing :)

JWo1F:
@rrousselGit, I've added a test for this behaviour.

JWo1F:
I increased the test coverage by adding a check to ensure that everything functions correctly after the value appears.

rrousselGit:
Looks like some files are improperly formatted. Could you format them?

Hooks use the master channel, so you might have to switch channel to use the latest formatter

JWo1F:
@rrousselGit, seems to be done now, could you check it, please?

rrousselGit:
Thanks!

rrousselGit:
Available in flutter_hooks 0.20.1

PR #383, opened 2023-09-27
feat(hook): implemented hook for MaterialStateController

AdamHavlicek:
this feature adds `useMaterialStatesController` hook with this we can change states of material 3 widgets


rrousselGit:
Hello!

This is looking good. Could you add this to the readme?

AdamHavlicek:
Yes, i gonna fix in additional commit. Do i have to create issue for this PR as well?
> This is looking good. Could you add this to the readme?

PR #394, opened 2023-11-13
Add hook useDeferredValue

itisnajim:
equivalent for react useDeferredValue hook.


rrousselGit:
Hello!

I'm not sure if that's really an equivalent to the React one.
I've never used it, but reading the docs, it seems like this deprioritise the subsequent render, such that it doesn't block the UI thread.
A simple setState after a timeout wouldn't do that.

packages/flutter_hooks/lib/src/deferred_value.dart
```diff
@@ -0,0 +1,74 @@
+part of 'hooks.dart';
+
+/// Returns a [ValueNotifier] that updates its value with a timeout
+/// for the given input [toDefer] and triggers widget rebuilds accordingly.
+ValueNotifier<T> useDeferredValue<T>(
+  T toDefer, {
+  Duration timeout = const Duration(seconds: 1),
```
rrousselGit, line 7: Where is that "one second" comes from? That delay looks arbitrary to me

itisnajim:
> Hello!
> 
> I'm not sure if that's really an equivalent to the React one. I've never used it, but reading the docs, it seems like this deprioritise the subsequent render, such that it doesn't block the UI thread. A simple setState after a timeout wouldn't do that.

yes it's not really equivalent to the react one, The background re-rendering process triggered by useDeferredValue happens as soon as React finishes the initial render, and the timeout can be set also, in react e.g:
```js
const deferredCount = useDeferredValue(count, {
    timeoutMs: 1000,
  });
```
in the code i've pushed it resembles more of a `debounce` mechanism, Maybe by incorporating  `WidgetsBinding.instance.addPostFrameCallback` we can ensure that the layout is completed.. still, I'm unsure about how to automatically manage and emit the `timeout` property in this context.

packages/flutter_hooks/lib/src/deferred_value.dart
itisnajim: maybe making `timeout` optional, if null or 0ms it could default to `300ms` and always combined with `WidgetsBinding.instance.addPostFrameCallback` then after that combination calling a setState to rebuild the widget and work with the deferred value ?

packages/flutter_hooks/lib/src/deferred_value.dart
```diff
@@ -0,0 +1,74 @@
...
+
+  @override
+  void initHook() {
+    super.initHook();
+    _startTimer(hook.toDefer);
+  }
+
+  void _startTimer(T toDefer) {
+    _timer?.cancel();
+    _timer = Timer(hook.timeout, () {
```
itisnajim, line 43: making `timeout` nullable and doing like this ?
```dart
_timer = Timer(hook.timeout ?? Duration(milliseconds: 300), () {
      WidgetsBinding.instance.addPostFrameCallback((_) {
        if (context.mounted) {
          _state.value = toDefer;
          setState(() {});
        }
      });
    });
```

rrousselGit:
As it is, wouldn't `usePrevious` be a good-enough replacement?

itisnajim:
usePrevious can't control the timeout meaning it doesn't wait for a specific duration. defer the update of a value for a specified timeout this can be beneficial for optimizing the rendering performance by delaying non-critical updates.
e.g: scenario where you have a search input component that triggers an API call for search suggestions. U can use useEffect with deferred.value as a dependency

```dart
useEffect(() {
  // Use deferredQuery for your API call
  fetchSuggestions(deferredQuery);

  // Cleanup function
  return () {};
}, [deferredQuery]);
```

packages/flutter_hooks/lib/src/deferred_value.dart
rrousselGit: We'd want a default similar to that of React, so definitely not 1 second.
If it's "right after the current render has completed", then we should do that too

rrousselGit:
I'm still skeptical whether that's really equivalent.

At the moment, the current implementation is only equivalent to a debounce/throttle. Yet the documentation of useDeferredValue is pretty clear in that it's different from a simple debounce/throttle

rrousselGit:
I would be more willing to add a simple "useDebounced" hook with an implementation similar to the current one.

This would avoid possible confusions in case someone familiar with the useDeferredValue hook tries to use flutter_hooks'

PR #395, opened 2023-11-15
Add hook useDebounced

itisnajim:
Following our earlier discussion in the PR at:
https://github.com/rrousselGit/flutter_hooks/pull/394#issuecomment-1812511330

it's clear that integrating the useDebounced hook will be quite beneficial!

### DIFF
--- a/packages/flutter_hooks/lib/src/debounced.dart
+++ b/packages/flutter_hooks/lib/src/debounced.dart
@@ -1,11 +1,27 @@
 part of 'hooks.dart';
 
-/// Returns a [ValueNotifier] that updates its value with a timeout
-/// for the given input [toDebounce] and triggers widget rebuilds accordingly.
-ValueNotifier<T?> useDebounced<T>(
-  T toDebounce, {
-  Duration timeout = const Duration(milliseconds: 500),
-}) {
+/// Returns a debounced version of the provided value [toDebounce], triggering
+/// widget updates accordingly after a specified [timeout] duration.
+///
+/// Example:
+/// ```dart
+/// String userInput = ''; // Your input value
+///
+/// // Create a debounced version of userInput
+/// final debouncedInput = useDebounced(
+///   userInput,
+///   Duration(milliseconds: 500), // Set your desired timeout
+/// );
+/// // Assume a fetch method fetchData(String query) exists
+/// useEffect(() {
+///   fetchData(debouncedInput); // Use debouncedInput as a dependency
+///   return null;
+/// }, [debouncedInput]);
+/// ```
+T? useDebounced<T>(
+  T toDebounce,
+  Duration timeout,
+) {
   return use(
     _DebouncedHook(
       toDebounce: toDebounce,
@@ -14,7 +30,7 @@ ValueNotifier<T?> useDebounced<T>(
   );
 }
 
-class _DebouncedHook<T> extends Hook<ValueNotifier<T?>> {
+class _DebouncedHook<T> extends Hook<T?> {
   const _DebouncedHook({
     required this.toDebounce,
     required this.timeout,
@@ -27,9 +43,8 @@ class _DebouncedHook<T> extends Hook<ValueNotifier<T?>> {
   _DebouncedHookState<T> createState() => _DebouncedHookState();
 }
 
-class _DebouncedHookState<T>
-    extends HookState<ValueNotifier<T?>, _DebouncedHook<T>> {
-  final ValueNotifier<T?> _state = ValueNotifier(null);
+class _DebouncedHookState<T> extends HookState<T?, _DebouncedHook<T>> {
+  T? _state;
   Timer? _timer;
 
   @override
@@ -41,25 +56,25 @@ class _DebouncedHookState<T>
   void _startDebounce(T toDebounce) {
     _timer?.cancel();
     _timer = Timer(hook.timeout, () {
-      if (context.mounted) {
-        _state.value = toDebounce;
-        setState(() {});
-      }
+      setState(() {
+        _state = toDebounce;
+      });
     });
   }
 
   @override
   void didUpdateHook(_DebouncedHook<T> oldHook) {
-    if (hook.toDebounce != oldHook.toDebounce) {
+    if (hook.toDebounce != oldHook.toDebounce ||
+        hook.timeout != oldHook.timeout) {
       _startDebounce(hook.toDebounce);
     }
   }
 
   @override
-  ValueNotifier<T?> build(BuildContext context) => _state;
+  T? build(BuildContext context) => _state;
 
   @override
-  Object? get debugValue => _state.value;
+  Object? get debugValue => _state;
 
   @override
   String get debugLabel => 'useDebounced<$T>';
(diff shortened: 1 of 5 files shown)


packages/flutter_hooks/lib/src/debounced.dart
```diff
@@ -0,0 +1,74 @@
+part of 'hooks.dart';
+
+/// Returns a [ValueNotifier] that updates its value with a timeout
+/// for the given input [toDebounce] and triggers widget rebuilds accordingly.
+ValueNotifier<T?> useDebounced<T>(
+  T toDebounce, {
+  Duration timeout = const Duration(milliseconds: 500),
```
rrousselGit, line 7: I would make the duration required and positional. For a useDebounced, that makes sense I think.

 I want to avoid arbitrary durations in the framework.

packages/flutter_hooks/lib/src/debounced.dart
```diff
@@ -0,0 +1,74 @@
+part of 'hooks.dart';
+
+/// Returns a [ValueNotifier] that updates its value with a timeout
+/// for the given input [toDebounce] and triggers widget rebuilds accordingly.
```
rrousselGit, line 4: Some examples and more docs would be great. For example we'd want to explain why we have a way to update the value

packages/flutter_hooks/lib/src/debounced.dart
```diff
@@ -0,0 +1,74 @@
...
+}) {
+  return use(
+    _DebouncedHook(
+      toDebounce: toDebounce,
+      timeout: timeout,
+    ),
+  );
+}
+
+class _DebouncedHook<T> extends Hook<ValueNotifier<T?>> {
```
rrousselGit, line 17: Also, why are we using T? instead of T?

packages/flutter_hooks/lib/src/debounced.dart
```diff
@@ -0,0 +1,74 @@
...
+
+  @override
+  Object? get debugValue => _state.value;
+
+  @override
+  String get debugLabel => 'useDebounced<$T>';
+
+  @override
+  void dispose() {
+    _timer?.cancel();
```
rrousselGit, line 69: We need to test that the timer is correctly disposed

packages/flutter_hooks/test/use_debounce_test.dart
```diff
@@ -0,0 +1,87 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_hooks/flutter_hooks.dart';
+import 'package:flutter_test/flutter_test.dart';
+
+void main() {
```
rrousselGit, line 6: We need to test that calling the state setter updates the UI

packages/flutter_hooks/lib/src/debounced.dart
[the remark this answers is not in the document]
rrousselGit: In fact, do we even need a `setValue`? Couldn't we return the value directly without a notifier?

packages/flutter_hooks/test/use_debounce_test.dart
```diff
@@ -0,0 +1,87 @@
...
+            ).value.toString(),
+            textDirection: TextDirection.ltr,
+          );
+        }),
+      );
+      await tester.pumpAndSettle(const Duration(milliseconds: 500));
+
+      expect(find.text('null'), findsOneWidget);
+    });
+    testWidgets('basic', (tester) async {
```
rrousselGit, line 49:
```suggestion

    testWidgets('basic', (tester) async {
```

packages/flutter_hooks/test/use_debounce_test.dart
```diff
@@ -0,0 +1,88 @@
...
+      ),
+    );
+  });
+
+  group('useDebounced', () {
+    testWidgets('default value is null', (tester) async {
+      await tester.pumpWidget(
+        HookBuilder(builder: (context) {
+          return Text(
+            useDebounced(
```
rrousselGit, line 37: Move this out of the Text(...) for readability

packages/flutter_hooks/lib/src/debounced.dart
```diff
@@ -0,0 +1,74 @@
...
+    _timer = Timer(hook.timeout, () {
+      if (context.mounted) {
+        _state.value = toDebounce;
+        setState(() {});
+      }
+    });
+  }
+
+  @override
+  void didUpdateHook(_DebouncedHook<T> oldHook) {
```
rrousselGit, line 52: Would we update the timer if the duration changes?

packages/flutter_hooks/lib/src/debounced.dart
```diff
@@ -0,0 +1,74 @@
...
+  @override
+  void initHook() {
+    super.initHook();
+    _startDebounce(hook.toDebounce);
+  }
+
+  void _startDebounce(T toDebounce) {
+    _timer?.cancel();
+    _timer = Timer(hook.timeout, () {
+      if (context.mounted) {
```
rrousselGit, line 44: We don't need to check `mounted` here.

packages/flutter_hooks/lib/src/debounced.dart
```diff
@@ -0,0 +1,74 @@
...
+        _state.value = toDebounce;
```
rrousselGit, line 45: This should be done inside `setState`

packages/flutter_hooks/lib/src/debounced.dart
```diff
@@ -0,0 +1,74 @@
...
+  Object? get debugValue => _state.value;
+
+  @override
+  String get debugLabel => 'useDebounced<$T>';
+
+  @override
+  void dispose() {
+    _timer?.cancel();
+    _timer = null;
+    _state.dispose();
```
rrousselGit, line 71: If we want to keep using a ValueNotifier, we need a test to verify that it is disposed

packages/flutter_hooks/lib/src/debounced.dart
itisnajim: If useDebounced is called for the first time and it's timer is still active, does that imply that there's no value yet (nothing debounced yet), making it null?

packages/flutter_hooks/lib/src/debounced.dart
itisnajim: i don't know how to do this! do i have to annotate timer with `@visibleForTesting` and then what ?

packages/flutter_hooks/lib/src/debounced.dart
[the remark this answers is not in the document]
itisnajim: `set` and `ValueNotifier` no needed (removed)

packages/flutter_hooks/lib/src/debounced.dart
rrousselGit: It's probably fine actually since Flutter tests will fail if there are unclosed timers

itisnajim:
I believe all should be well at this point
* Using `timeout` as a positional argument instead of named.
* Adding more comments/docs.
* Removing `setValue()` as its unnecessary.
* Double checking and fixing tests.
https://github.com/rrousselGit/flutter_hooks/pull/395/commits/0d3f71f101ef818533874ef8fd19aa0720f2c218

rrousselGit:
Looking good, Thanks!

Issue #132, opened 2020-05-07
Clarify useFuture usage


fgblomqvist:
In #119 you mentioned that you are expected/have to use `useMemoized` with `useFuture` to prevent it from getting called multiple times. While I can sort of understand that, it's not made clear in the docs. More importantly, is there ever a time when you would want to `useFuture` without `useMemoized`? It seems to be that the most common usage would require `useMemoized`, and in that case, I feel like `useFuture` should have that "built-in" so-to-speak.

fgblomqvist:
I ended up writing my own hook that combines `useFuture` and `useMemoized`, and also offers a way to refresh the future in an elegant fashion:

```dart
import 'package:flutter/widgets.dart';
import 'package:flutter_hooks/flutter_hooks.dart';

/// Stores an [AsyncSnapshot] as well as a reference to a function [refresh]
/// that should re-call the future that was used to generate the [snapshot].
class MemoizedAsyncSnapshot<T> {
  final AsyncSnapshot<T> snapshot;
  final Function() refresh;

  const MemoizedAsyncSnapshot(this.snapshot, this.refresh);
}

/// Subscribes to a [Future] and returns its current state in a
/// [MemoizedAsyncSnapshot].
/// The [future] is memoized and will only be re-called if any of the [keys]
/// change or if [MemoizedAsyncSnapshot.refresh] is called.
///
/// * [initialData] specifies what initial value the [AsyncSnapshot] should
///   have.
/// * [preserveState] defines if the current value should be preserved when
///   changing the [Future] instance.
///
/// See also:
///   * [useFuture], the hook responsible for getting the future.
///   * [useMemoized], the hook responsible for the memoization.
MemoizedAsyncSnapshot<T> useMemoizedFuture<T>(
  Future<T> Function() future, {
  List<Object> keys = const <dynamic>[],
  T initialData,
  bool preserveState = true,
}) {
  final refresh = useState(0);
  final result = useFuture(
    useMemoized(future, [refresh.value, ...keys]),
    initialData: initialData,
    preserveState: preserveState,
  );

  void refreshMe() => refresh.value++;

  return MemoizedAsyncSnapshot<T>(result, refreshMe);
}
```
I think this hook is the missing piece in the puzzle that is Flutter and memoization of futures (regardless of whether you use hooks or not, there doesn't seem to be a good pattern on how to manage refreshing of your futures?).

Here's how I use it (in a basic fashion):
```dart
[...]

Issue #119, opened 2020-01-03
`useFuture` is triggered multiple times.


zaralockheart:
A snipet
```
import 'package:flutter/material.dart';
import 'package:flutter_hooks/flutter_hooks.dart';

class MyWidget extends HookWidget {
  @override
  Widget build(BuildContext context) {
    final myFuture = useFuture(
        Future.delayed(Duration(microseconds: 100), () => print("TESTTT")));

    return Scaffold(
      body: Container(),
    );
  }
}
```

When I run this code. Log in console give me this.
```
Performing hot restart...
Syncing files to device Android SDK built for x86...
Restarted application in 3,573ms.
I/flutter (21643): TESTTT
I/flutter (21643): TESTTT
I/flutter (21643): TESTTT
I/flutter (21643): TESTTT
I/flutter (21643): TESTTT
I/flutter (21643): TESTTT
I/flutter (21643): TESTTT
I/flutter (21643): TESTTT
I/flutter (21643): TESTTT
I/flutter (21643): TESTTT
I/flutter (21643): TESTTT
I/flutter (21643): TESTTT
I/flutter (21643): TESTTT
```

Is this expected? Are there any ways to make it trigger once?

rrousselGit:
This is expected. You should cache the future in a `useMemoized` first

zaralockheart:
thanks for quick reply! also thank you for the answer :)

yasirhantoush:
a solution that worked for me as a reference:

```
import 'package:flutter/material.dart';
import 'package:flutter_hooks/flutter_hooks.dart';

class MyWidget extends HookWidget {
  @override
  Widget build(BuildContext context) {
    final retryCount = useState(0); // used if you want to refresh the future
    final myFuture = useMemoized(() { 
        return Duration(microseconds: 100), () => print("TESTTT"));
    }, [retryCount]);

    final mySnapShot = useFuture(myFuture);

    return Scaffold(
      body: Container(),
    );
  }
}
```

PR #408, opened 2024-01-04
Clarify the use of useFuture

Region40:
**Describe what scenario you think is uncovered by the existing examples/articles**
Document that Futures should not be created inside `useFuture` (see #132 & #119)

**Describe why existing examples/articles do not cover this case**
`useFuture` behaves like `FutureBuilder`. This is not explicitly stated but rather implied through an understanding of Flutter Hooks. This understanding may not be intuitive for new users and can lead to a misunderstanding about the behavior of `useFuture`.


rrousselGit:
Thanks!

PR #411, opened 2024-01-16
Upgrade to Flutter 3.7.0 and Dart 2.19

Reprevise:
Prerequisite to deprecate the `useIsMounted` hook which will be done in a separate PR so this is kept clean to resolve any issues there are during upgrading. I updated the changelog as well, and all tests are passing.

### DIFF
--- a/packages/flutter_hooks/CHANGELOG.md
+++ b/packages/flutter_hooks/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.21.0 - unreleased
+
+- Upgrade Flutter to 3.7.0 and Dart 2.19.0 (thanks to @Reprevise)
+
 ## 0.20.4 - 2023-12-29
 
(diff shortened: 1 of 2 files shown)


rrousselGit:
I don't think that's necessary. We could deprecate the hook without bumping minimum Flutter

rrousselGit:
That doesn't matter.

Reprevise:
I'm confused as to why it doesn't matter. This hook should only be deprecated when users have a valid alternative. To ensure that they do, we need to bump the minimum flutter version so users who are below it can't upgrade to the new version of the package.

rrousselGit:
The hook is deprecated regardless of whether people use Flutter 3.7.0 or not.
The only thing this would do is prevent folks using an out of date SDK from knowing that they use a deprecated API.

PR #415, opened 2024-02-12
Reduce boilerplate code of controller (ChangeNotifier) hooks

yelliver:
If this PR is merged, I will group all of them into a single file, making the repo cleaner.

### DIFF
--- a/packages/flutter_hooks/lib/src/change_notifier.dart
+++ b/packages/flutter_hooks/lib/src/change_notifier.dart
@@ -3,18 +3,13 @@ part of 'hooks.dart';
 /// Creates a general purpose hook on a [ChangeNotifier] subclass that will be
 /// disposed automatically.
 ///
-/// For example, these are equivalent:
+/// For example, this hook:
 /// ```dart
 /// useTextEditingController(text: 'Hello');
 /// ```
-/// And:
+/// which is actually implemented by using `useChangeNotifier`:
 /// ```dart
 /// useChangeNotifier(() => TextEditingController(text: 'Hello'));
-/// ```
-/// This is very useful if you can't find a "common controller" class in Flutter
-/// which has not a hook implementation, for example `CupertinoTabController`:
-/// ```dart
-/// useChangeNotifier(() => CupertinoTabController());
 /// ```
 T useChangeNotifier<T extends ChangeNotifier>(
   T Function() factory, [
--- a/packages/flutter_hooks/lib/src/change_notifier.dart
+++ b/packages/flutter_hooks/lib/src/change_notifier.dart
@@ -9,12 +9,12 @@ part of 'hooks.dart';
 /// ```
 /// And:
 /// ```dart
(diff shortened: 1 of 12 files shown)


rrousselGit:
That looks interesting.

A few things:
- We'd need tests for the new hooks
- They should be added to the readme and the changelog

packages/flutter_hooks/lib/src/change_notifier.dart
```diff
@@ -0,0 +1,30 @@
...
+/// useDisposable(() => TextEditingController(text: 'Hello'));
+/// ```
+/// This is very useful if you can't find a "common controller" class in Flutter
+/// which has not a hook implementation, for example `CupertinoTabController`:
+/// ```dart
+/// useDisposable(() => CupertinoTabController());
+/// ```
+T useChangeNotifier<T extends ChangeNotifier>(
+  T Function() factory, [
+  String? label,
```
rrousselGit, line 21: I don't like this being part of the public API

yelliver:
> That looks interesting.
> 
> A few things:
> 
> * We'd need tests for the new hooks
> * They should be added to the readme and the changelog

- I think the new hooks are used by existing hooks (which are changed in implementation) so their existing test cases should cover, how do you think?
- If we don't expose `useChangeNotifier` & `useDispose`, do we need any info in readme?

packages/flutter_hooks/lib/src/change_notifier.dart
yelliver: as I commented, it can be used to hook a common ChangeNotifier that has not hook yet, for example `CupertinoTabController` which was expected by some people: `[CupertinoTabController](https://api.flutter.dev/flutter/cupertino/CupertinoTabController-class.html)`

rrousselGit:
> If we don't expose useChangeNotifier & useDispose, do we need any info in readme?

No but then we should mark those as `@internal`

rrousselGit:
> I think the new hooks are used by existing hooks (which are changed in implementation) so their existing test cases should cover, how do you think?

Not if we export those hooks.

rrousselGit:
:wave:
Do you still wish to work on this? If so, we need to decide wether you want those hooks part of the public API, or be internal only (and therefore not exported).

yelliver:
Let's recap:
- If we want to export them -> Need to add test cases for them
- If we don't export them -> Make them @internal

I prefer to export them, if you think it is proper, I will add test cases for them.

rrousselGit:
If we want to export them, I'll also be a bit more strict on the API.

yelliver:
which restriction do you want to change? I will change it

lvsecoto:
I think we need to export them, to reduce these boring code:
```dart
      final controller = useMemoized(
        () => SomeChangeNotifier(),
      );
      useEffect(
        () => () {
          controller.dispose();
        },
        [controller],
      );
```

rrousselGit:
This appears to be stale, so I'll close this :)

Feel free to reopen a PR if you want to resume working on this

PR #421, opened 2024-03-18
Adds check for either WidgetStates or MaterialStates in useMaterialStatesController debug test

MitchellGoodwin:
Flutter is moving MaterialState, and it's related functions outside of the Material library and changing it's name to WidgetState. MaterialState will still be available as is, just marked deprecated, and made a typedef of WidgetState. However this causes MaterialState debug functions to return with strings mentioning "WidgetState", and is a breaking change for packages using those functions. The [PR that will add that change to Flutter is here](https://github.com/flutter/flutter/pull/142151), and the [original issue is here](https://github.com/flutter/flutter/issues/138270).

Internal testing found this caused a breakage in this package, so this PR updates the test for `useMaterialStatesController` to pass both before and after that change is landed.

### DIFF
--- a/packages/flutter_hooks/test/use_material_states_controller_test.dart
+++ b/packages/flutter_hooks/test/use_material_states_controller_test.dart
@@ -21,10 +21,17 @@ void main() {
       element
           .toDiagnosticsNode(style: DiagnosticsTreeStyle.offstage)
           .toStringDeep(),
-      equalsIgnoringHashCodes(
-        'HookBuilder\n'
-        ' │ useMaterialStatesController: MaterialStatesController#00000({})\n'
-        ' └SizedBox(renderObject: RenderConstrainedBox#00000)\n',
+      anyOf(
+        equalsIgnoringHashCodes(
+          'HookBuilder\n'
+          ' │ useMaterialStatesController: MaterialStatesController#00000({})\n'
+          ' └SizedBox(renderObject: RenderConstrainedBox#00000)\n',
+        ),
+        equalsIgnoringHashCodes(
+          'HookBuilder\n'
+          ' │ useMaterialStatesController: WidgetStatesController#00000({})\n'
+          ' └SizedBox(renderObject: RenderConstrainedBox#00000)\n',
+        )
       ),
     );
   });
--- a/packages/flutter_hooks/test/use_material_states_controller_test.dart
+++ b/packages/flutter_hooks/test/use_material_states_controller_test.dart
@@ -22,17 +22,16 @@ void main() {
           .toDiagnosticsNode(style: DiagnosticsTreeStyle.offstage)
           .toStringDeep(),
       anyOf(
-        equalsIgnoringHashCodes(
-          'HookBuilder\n'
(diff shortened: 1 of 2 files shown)


packages/flutter_hooks/test/use_material_states_controller_test.dart
```diff
@@ -21,10 +21,17 @@ void main() {
...
-        ' └SizedBox(renderObject: RenderConstrainedBox#00000)\n',
+      anyOf(
+        equalsIgnoringHashCodes(
+          'HookBuilder\n'
+          ' │ useMaterialStatesController: MaterialStatesController#00000({})\n'
+          ' └SizedBox(renderObject: RenderConstrainedBox#00000)\n',
+        ),
+        equalsIgnoringHashCodes(
+          'HookBuilder\n'
+          ' │ useMaterialStatesController: WidgetStatesController#00000({})\n'
```
rrousselGit, line 32: This feels a bit weird to me. `useMaterialStatesController` is explicitly instantiating `MaterialStatesController`

How did we end-up with a `WidgetStatesController`? Wouldn't we get a type error?

rrousselGit: Oh `MaterialStatesController` is deprecated and it's a typedef

rrousselGit:
Considering MaterialStatesController is deprecated, the failing "hooks" should be deprecated too.

Does this PR have to land before https://github.com/rrousselGit/flutter_hooks/pull/421?

I'd rather:
- wait for https://github.com/rrousselGit/flutter_hooks/pull/421
- deprecate useMaterialStatesController
- add a `useWidgetStatesController` using non-deprecated API
- only use `WidgetStatesController` in the test (rather than have the test depend on deprecated API)

packages/flutter_hooks/test/use_material_states_controller_test.dart
MitchellGoodwin: It is weird. `MaterialStatesController` is being moved down to the widgets layer and renamed to `WidgetStatesController`.  For migration purposes, there is still a `MaterialStatesController`, but it is just a typedef of `WidgetStatesController`, so it acts the same as how it did before. Except for the debug functions, which return strings mentioning `WidgetState`.

MitchellGoodwin:
> Does this PR have to land before https://github.com/rrousselGit/flutter_hooks/pull/421?

I saw that the Flutter PR broke the test in this package, so I wanted to add a fix for the test that would pass both before and after the change in Flutter. Theoretically #421 can land before [flutter/142151](https://github.com/flutter/flutter/pull/142151), then a `useWidgetStatesController` could be added anytime before `MaterialStatesController` is removed after being marked deprecated.

But if you'd rather not adjust the upcoming broken test and add a `useWidgetStatesController` right away, then that's fine too.

rrousselGit:
Awesome. I'd rather things are they are right now then.
The failing test will serve as a reminder to switch to the new API once the CI fails :)

Thanks a lot for the heads up though!

MitchellGoodwin:
No problem. I'll close the PR 👍

PR #422, opened 2024-03-19
Update `SearchAnchor` test for pending timer

TahaTesser:
This PR fixes the pending timer when `SearchAnchor` is opened in the "check opening/closing view" test in the `use_search_controller_test.dart` class.

Unblocks a [Flutter bug fix](https://github.com/flutter/flutter/pull/143443#issuecomment-2004655552) as `flutter_hooks` tests are part of the Google Testing suite.

### DIFF
--- a/packages/flutter_hooks/test/use_search_controller_test.dart
+++ b/packages/flutter_hooks/test/use_search_controller_test.dart
@@ -71,6 +71,8 @@ void main() {
       controller.openView();
 
       expect(controller.isOpen, true);
+      await tester.pump();
+      await tester.pump(const Duration(seconds: 150)); // Advance fade animation duration.
 
       await tester.pumpWidget(MaterialApp(
         home: HookBuilder(builder: (context) {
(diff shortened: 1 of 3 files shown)


packages/flutter_hooks/test/use_search_controller_test.dart
```diff
@@ -71,6 +71,8 @@ void main() {
       controller.openView();
 
       expect(controller.isOpen, true);
+       // Advance fade animation duration.
+      await tester.pumpAndSettle(const Duration(seconds: 150));
```
rrousselGit, line 75: I don't like this change. I'm not testing animations here. Why would this be needed?

TahaTesser:
> I don't like this change. I'm not testing animations here. Why would this be needed?

Me too. I didn't see another way but thanks to your suggestion https://github.com/flutter/flutter/pull/143443, the change in the Flutter is no longer the failing the `flutter_hooks` test, closing as a result.

PR #438, opened 2024-07-22
Add `useOnListenableChange`

### LINKED ISSUE #326: Trigger a callback when value listenable changes
**Is your feature request related to a problem? Please describe.**
Having a big and complicated screen, that is optimised using `useValueNotifier` instead of `useState`, often there is a need to do something when value notifier changes.
Using `useValueListenable` would trigger rebuild, which is unwanted due to rebuilds optimisation

**Describe the solution you'd like**
There is a way to add and remove listener to value notifier. This could be utilised to build a `useValueListenableChanged` hook, similar to `useValueChanged` hook that is present already.

**Describe alternatives you've considered**
`useValueChanged` hook works on a regular variable, not on `valueListenable`

**Additional context**
Here is what seemed to have worked. If there is no negative feedback I can make a PR:

```
import 'package:flutter/foundation.dart';
import 'package:flutter/src/widgets/framework.dart';
import 'package:flutter_hooks/flutter_hooks.dart';

useValueListenableChanged<T>(
    ValueListenable<T> listenable, Function(T value) valueListenableChanged) {
  return use(
      _UseValueListenableChangedHook(listenable, valueListenableChanged));
}

class _UseValueListenableChangedHook<T> extends Hook {
  const _UseValueListenableChangedHook(
      this.valueListenable, this.valueListenableChanged);

  final Function(T value) valueListenableChanged;
  final ValueListenable<T> valueListenable;

  @override
  _ValueListenableChangedState createState() =>
      _ValueListenableChangedState<T>();
}

class _ValueListenableChangedState<T>
    extends HookState<dynamic, _UseValueListenableChangedHook<T>> {
  _ValueListenableChangedState();

  late VoidCallback _valueListenableChanged;

  @override
  void initHook() {
    _valueListenableChanged = () {
      hook.valueListenableChanged(hook.valueListenable.value);
    };
    hook.valueListenable.addListener(_valueListenableChanged);
  }

  @override
  void dispose() {
    hook.valueListenable.removeListener(_valueListenableChanged);
  }

  @override
  build(BuildContext context) {}

  @override
  String get debugLabel => 'useValueListenableChanged';

  @override
  Object? get debugValue => hook.valueListenable;
}

```

### LINKED ISSUE #389: Feature Request: `useListener` hook
**Is your feature request related to a problem? Please describe.**

When using `Listenable` with `useEffect` it can be a bit cumbersome to always repeat the same boilerplate off `addListener` and `removeListener` again and again:

* Having to call `keys.length` times `addListener`
* Duplication of key in keys and `key.addListener(...)` for each key
* Extra level of indentation when writing the listener inside the `useEffect` closure

**Describe the solution you'd like**

Add a `useListener` hook.

```dart
void useListener(void Function() listener, List<Listenable> listenables) {
  useEffect(() {
    final listenable = Listenable.merge(listenables);
    listenable.addListener(listener);

    return () => listenable.removeListener(listener);
  }, listenables);
}
```

**Additional context**

This would be very useful to bind values when some controller changes:

```dart
final nameError = useState<String?>(null);
final nameController = useTextEditingController();

useListener(() {
  final validation = nameValidation(offlineNameController.text);
  nameError.value = validation.nameError;
}, [nameController]);
```

and even when it depends on multiple controllers like:

```dart
final activeTab = useState(_initialTab);
final primaryTabController = useTabController(
  initialLength: Tabs.primaryTabLength,
  initialIndex: activeTab.value.primaryTabIndex,
);
final secondaryTabController = useTabController(
  initialLength: Tabs.secondaryTabLength,
  initialIndex: activeTab.value.secondaryTabIndex,
);

useListener(() {
  activeTab.value = Tabs.of(
    primaryTabController.index,
    secondaryTabController.index,
  );
}, [
  primaryTabController,
  secondaryTabController,
]);
```

timcreatedit:
Fixes #326
(potentially, people can just call `.value` for `ValueListenables`, so I would argue an extra hook isn't needed
Fixes #389

### DIFF
--- a/packages/flutter_hooks/lib/src/listenable.dart
+++ b/packages/flutter_hooks/lib/src/listenable.dart
@@ -151,9 +151,8 @@ void useOnListenableChange(
 class _OnListenableChangeHook extends Hook<void> {
   const _OnListenableChangeHook(
     this.listenable,
-    this.listener, {
-    super.keys,
-  });
+    this.listener,
+  );
 
   final Listenable? listenable;
   final VoidCallback listener;
--- a/packages/flutter_hooks/lib/src/listenable.dart
+++ b/packages/flutter_hooks/lib/src/listenable.dart
@@ -165,16 +165,15 @@ class _OnListenableChangeHookState
   @override
   void initHook() {
     super.initHook();
-    hook.listenable?.addListener(hook.listener);
+    hook.listenable?.addListener(_listener);
   }
 
   @override
   void didUpdateHook(_OnListenableChangeHook oldHook) {
     super.didUpdateHook(oldHook);
-    if (hook.listenable != oldHook.listenable ||
-        oldHook.listener != hook.listener) {
-      oldHook.listenable?.removeListener(oldHook.listener);
-      hook.listenable?.addListener(hook.listener);
+    if (hook.listenable != oldHook.listenable) {
+      oldHook.listenable?.removeListener(_listener);
+      hook.listenable?.addListener(_listener);
     }
   }
 
@@ -184,6 +183,11 @@ class _OnListenableChangeHookState
   @override
   void dispose() {
     hook.listenable?.removeListener(hook.listener);
+  }
+
+  /// Wraps `hook.listener` so we have a non-changing reference to it.
+  void _listener() {
+    hook.listener();
   }
 
   @override
--- a/packages/flutter_hooks/lib/src/listenable.dart
+++ b/packages/flutter_hooks/lib/src/listenable.dart
@@ -142,10 +142,9 @@ class _UseValueNotifierHookState<T>
 ///  * [useListenable]
 void useOnListenableChange(
   Listenable? listenable,
-  VoidCallback listener, {
-  List<Object?>? keys,
-}) {
-  return use(_OnListenableChangeHook(listenable, listener, keys: keys));
+  VoidCallback listener,
+) {
+  return use(_OnListenableChangeHook(listenable, listener));
 }
 
 class _OnListenableChangeHook extends Hook<void> {
@@ -172,7 +171,8 @@ class _OnListenableChangeHookState
   @override
   void didUpdateHook(_OnListenableChangeHook oldHook) {
     super.didUpdateHook(oldHook);
-    if (hook.listenable != oldHook.listenable) {
+    if (hook.listenable != oldHook.listenable ||
+        oldHook.listener != hook.listener) {
       oldHook.listenable?.removeListener(oldHook.listener);
       hook.listenable?.addListener(hook.listener);
     }
--- a/packages/flutter_hooks/lib/src/listenable.dart
+++ b/packages/flutter_hooks/lib/src/listenable.dart
@@ -128,3 +128,68 @@ class _UseValueNotifierHookState<T>
   @override
   String get debugLabel => 'useValueNotifier';
 }
+
+/// Adds a given [listener] to a [Listenable] and removes it when the hook is
+/// disposed.
+///
+/// As opposed to `useListenable`, this hook does not mark the widget as needing
+/// build when the listener is called. Use this for side effects that do not
+/// require a rebuild.
+///
+/// See also:
+///  * [Listenable]
+///  * [ValueListenable]
+///  * [useListenable]
+void useOnListenableChange<T extends Listenable>(
+  T? listenable,
+  VoidCallback listener, {
+  List<Object?>? keys,
+}) {
+  return use(_OnListenableChangeHook(listenable, listener, keys: keys));
+}
+
+class _OnListenableChangeHook extends Hook<void> {
+  const _OnListenableChangeHook(
+    this.listenable,
+    this.listener, {
+    super.keys,
+  });
+
+  final Listenable? listenable;
+  final VoidCallback listener;
+
+  @override
+  _OnListenableChangeHookState createState() => _OnListenableChangeHookState();
+}
+
+class _OnListenableChangeHookState
+    extends HookState<void, _OnListenableChangeHook> {
+  @override
+  void initHook() {
+    super.initHook();
+    hook.listenable?.addListener(hook.listener);
+  }
+
+  @override
+  void didUpdateHook(_OnListenableChangeHook oldHook) {
+    super.didUpdateHook(oldHook);
+    if (hook.listenable != oldHook.listenable) {
+      oldHook.listenable?.removeListener(oldHook.listener);
+      hook.listenable?.addListener(hook.listener);
+    }
+  }
+
+  @override
+  void build(BuildContext context) {}
+
+  @override
+  void dispose() {
+    hook.listenable?.removeListener(hook.listener);
+  }
+
+  @override
+  String get debugLabel => 'useOnListenableChange';
+
+  @override
+  Object? get debugValue => hook.listenable;
+}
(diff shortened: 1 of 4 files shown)


packages/flutter_hooks/lib/src/listenable.dart
```diff
@@ -128,3 +128,68 @@ class _UseValueNotifierHookState<T>
...
+///
+/// As opposed to `useListenable`, this hook does not mark the widget as needing
+/// build when the listener is called. Use this for side effects that do not
+/// require a rebuild.
+///
+/// See also:
+///  * [Listenable]
+///  * [ValueListenable]
+///  * [useListenable]
+void useOnListenableChange<T extends Listenable>(
```
rrousselGit, line 143: There's no need to make it generic

timcreatedit: True ofc, fixed!

packages/flutter_hooks/lib/src/listenable.dart
```diff
@@ -128,3 +128,68 @@ class _UseValueNotifierHookState<T>
...
+/// require a rebuild.
+///
+/// See also:
+///  * [Listenable]
+///  * [ValueListenable]
+///  * [useListenable]
+void useOnListenableChange(
+  Listenable? listenable,
+  VoidCallback listener, {
+  List<Object?>? keys,
```
rrousselGit, line 146: Keys aren't necessary.

And we should naturally support `listener` changing.

packages/flutter_hooks/test/use_on_listenable_change_test.dart
```diff
@@ -0,0 +1,110 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_hooks/flutter_hooks.dart';
+
+import 'mock.dart';
+
+void main() {
+  testWidgets('debugFillProperties', (tester) async {
```
rrousselGit, line 8: We need a test for checking that the listener changed

packages/flutter_hooks/lib/src/listenable.dart
timcreatedit: Good catch, oversight on my end

packages/flutter_hooks/lib/src/listenable.dart
```diff
@@ -173,7 +171,8 @@ class _OnListenableChangeHookState
   @override
   void didUpdateHook(_OnListenableChangeHook oldHook) {
     super.didUpdateHook(oldHook);
-    if (hook.listenable != oldHook.listenable) {
+    if (hook.listenable != oldHook.listenable ||
```
rrousselGit, line 174: That's not very efficient. It'd usually add/remove a listener whenever the build method updates.
We should have a `listener` method on the Hook class.

Of course we'd need a test for that too :)

timcreatedit: Hmm you're right, I never thought about that, most people are going to use it with anonymous functions. Time to refactor all my codebases that use callbacks in hooks as well 😅

timcreatedit: I've updated it, I'm not quite sure if this is what you meant, but it seems to work

rrousselGit: The issue isn't the anonymous function. It's your didUpdateHook implementation.

Like I said, you could have:

```dart
class A extends Hook {
  void listener() {
    hook.listener();
  }

  initHook() {
    listenable.addListener(listener);
  }
  dispose() {
    listenable.removeListener(listener);
  }
}
```

timcreatedit: Thanks for the feedback, however, I honestly can't quite make sense of what you're saying.

Both your previous comment, as well as this one refer to having a `listener` method on the hook class, however your pseudo-example calls `hook.listener()` from inside a `Hook` not a `HookState`, which must be a mistake. 

I'm therefore assuming you meant to write `extends HookState`. It makes sense to optimize `listener` changes like this, which is what I did on the most recent commits, but I'm failing to see how this would update whenever the Hook's `listenable` changes?

packages/flutter_hooks/test/use_on_listenable_change_test.dart
```diff
@@ -0,0 +1,171 @@
...
+    );
+
+    expect(value, isNull);
+    listenable.value++;
+    expect(value, 43);
+  });
+
+  testWidgets(
+    'listens new Listenable when Listenable is changed',
+    (tester) => tester.runAsync(() async {
```
rrousselGit, line 56: Those runAsync aren't necessary

packages/flutter_hooks/test/use_on_listenable_change_test.dart
```diff
@@ -0,0 +1,171 @@
...
+          builder: (context) {
+            useOnListenableChange(listenable, listener1);
+            return const SizedBox();
+          },
+        ),
+      );
+
+      await tester.pumpWidget(
+        HookBuilder(
+          key: const Key('hook_builder'),
```
rrousselGit, line 113: No need for keys

packages/flutter_hooks/test/use_on_listenable_change_test.dart
```diff
@@ -0,0 +1,171 @@
...
+        return const SizedBox();
+      }),
+    );
+
+    // ignore: invalid_use_of_protected_member
+    expect(listenable.hasListeners, isTrue);
+
+    await tester.pumpWidget(
+      Container(),
+    );
```
rrousselGit, line 166:
```suggestion
    await tester.pumpWidget(Container());
```

packages/flutter_hooks/lib/src/listenable.dart
rrousselGit: Tests are still passing, aren't they? ;)

`this.hook` was mutated when `didUpdateHook` was called. So the `HookState.listener` method points to the new callback.

timcreatedit: I think there's a misunderstanding. If you agree with the latest implementation on this branch, we're on the same page. I was under the assumption that you want to get rid of the entire `didUpdateHook` method as well. That's the part that doesn't make sense to me.

timcreatedit: If the only parts where the listener is added and removed are `initHook` and `dispose` respectively, which is what I thought you were pointing at, tests don't pass anymore.

packages/flutter_hooks/test/use_on_listenable_change_test.dart
timcreatedit: Caught me copy-pasting from onStreamChanged tests there 😅

timcreatedit:
@rrousselGit, I was wondering whether we were on the same page now or if you expect me to do anything still? Just wanted to touch base since I don't know if, and which changes you are proposing, so I'm currently not working on anything 😅

rrousselGit (approved):
Sorry, my bad.

LGTM, but do you mind adding it in the README?

timcreatedit:
> LGTM, but do you mind adding it in the README?

It's added to the "Listenable related hooks:" table, would you want it somewhere else?

rrousselGit:
Ah dang it, I didn't see the file.
It's good. Thanks!

PR #440, opened 2024-08-04
Reduce boilerplate code in `Change Notifier`-based hooks

yelliver:
This PR is a continuation of PR#123 and ensures that:

- No public API is changed or affected
- No new API is published

I want to keep this PR focused on reducing boilerplate code and do not intend to add a complex hook with various life-cycles (this might be done in a separate PR).


yelliver:
@rrousselGit could u pls take a look?

rrousselGit:
I appreciate, but I'd rather stick to the current implementation.

The mains are minimal, and we loose a bit in terms of diagnostic and debugging. Using a unique class for all hooks produce better error messages and improves hot-reload a bit.

PR #456, opened 2025-01-28
Add weak property to useState

westito:
I missed a function that can easily update value in `useState` when initial data changes, without using additional hooks like `useEffect` or `onValueChanged`. It is very common in Flutter when using widgets.

## Summary by CodeRabbit

- **New Features**
	- Enhanced `useState` hook with a new `weak` parameter
	- Added more flexible state management control in Flutter hooks
	- Introduced ability to conditionally update state across widget rebuilds

- **Tests**
	- Added comprehensive test cases for new `useState` hook behavior
	- Verified state update scenarios with weak and standard state management


rrousselGit:
I appreciate, but I'd rather not add parameters to core hooks that are not present in React.

westito:
Alternative implementation could be a separate "useWeakState" hook?

rrousselGit:
I don't think a lot of people would use it.

For now I'd suggest adding that in your app. Not all hooks need to be in flutter_hooks itself :)

PR #463, opened 2025-03-17
Expanding the documentation

nate-thegrate:
I've been doing things with Hooks a lot over the past few months, and it's been really fun to learn about the inner mechanisms of flutter_hooks.

The goal of this PR is to flesh out a few of the API descriptions and to tweak the wording in a few places for better accuracy.

## Summary by CodeRabbit

- **Documentation**
  - Clarified Hook usage: how widgets use hooks to manage mutable data and that hooks must be called unconditionally and in a fixed order during build.
  - Explained when rebuild checks are skipped and how cached hook values are used to avoid unnecessary rebuilds.
  - Clarified the roles and lifecycle of HookState, HookElement, HookWidget, and StatefulHookWidget.
  - Improved wording around context access and proper usage within build.

### DIFF
--- a/packages/flutter_hooks/lib/src/framework.dart
+++ b/packages/flutter_hooks/lib/src/framework.dart
@@ -17,13 +17,19 @@ bool debugHotReloadHooksEnabled = true;
 // ignore: deprecated_member_use, deprecated_member_use_from_same_package
 R use<R>(Hook<R> hook) => Hook.use(hook);
 
-/// [Hook] is similar to a [StatelessWidget], but is not associated
-/// to an [Element].
+/// Allows a [Widget] to create and access its own mutable data
+/// without implementing a [State].
 ///
-/// A [Hook] is typically the equivalent of [State] for [StatefulWidget],
-/// with the notable difference that a [HookWidget] can have more than one [Hook].
-/// A [Hook] is created within the [HookState.build] method of a [HookWidget] and the creation
-/// must be made unconditionally, always in the same order.
+/// Whereas [Widget]s store the immutable configuration for UI components,
+/// [Hook]s store immutable configuration for any type of object.
+/// The [HookState] of a [Hook] is analogous to the [State] of a [StatefulWidget],
+/// and a single [HookWidget] can use more than one [Hook].
+///
+/// Hooks can be used by replacing `extends StatelessWidget` with `extends HookWidget`,
+/// or by replacing `Builder()` with `HookBuilder()`.
+///
+/// Hook functions must be called unconditionally during the `build()` method,
+/// and always in the same order.
 ///
 /// ### Good:
 /// ```
@@ -216,7 +222,11 @@ Calling them outside of build method leads to an unstable state and is therefore
   HookState<R, Hook<R>> createState();
 }
 
-/// The logic and internal state for a [HookWidget]
+/// The logic and internal state of a [Hook].
+///
+/// This class is similar to a [State], but instead of building a [Widget]
+/// subtree, the [build] method can return a value of any type, as specified
+/// by the "result" type argument `R`.
 abstract class HookState<R, T extends Hook<R>> with Diagnosticable {
   /// Equivalent of [State.context] for [HookState]
   @protected
@@ -277,22 +287,19 @@ abstract class HookState<R, T extends Hook<R>> with Diagnosticable {
 
   /// Called before a [build] triggered by [markMayNeedRebuild].
   ///
-  /// If [shouldRebuild] returns `false` on all the hooks that called [markMayNeedRebuild]
-  /// then this aborts the rebuild of the associated [HookWidget].
+  /// If [shouldRebuild] returns `false` on all the hooks that called [markMayNeedRebuild],
+  /// [HookElement.build] will return a cached value instead of rebuilding each [Hook].
   ///
-  /// There is no guarantee that this method will be called after [markMayNeedRebuild]
-  /// was called.
-  /// Some situations where [shouldRebuild] will not be called:
-  ///
-  /// - [setState] was called
-  /// - a previous hook's [shouldRebuild] returned `true`
-  /// - the associated [HookWidget] changed.
+  /// This method is not evaluated if a previous Hook called [markMayNeedRebuild]
+  /// and its [shouldRebuild] method returned `true`.
+  /// Additionally, if [setState], [didUpdateHook], or [HookElement.didChangeDependencies] is called,
+  /// the build is unconditional and the `shouldRebuild()` call is skipped.
   bool shouldRebuild() => true;
 
   /// Mark the associated [HookWidget] as **potentially** needing to rebuild.
   ///
   /// As opposed to [setState], the rebuild is optional and can be cancelled right
-  /// before `build` is called, by having [shouldRebuild] return false.
+  /// before [build] is called, by having [shouldRebuild] return false.
   void markMayNeedRebuild() {
     if (_element!._isOptionalRebuild != false) {
       _element!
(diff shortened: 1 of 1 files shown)


packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -10,20 +10,27 @@ bool debugHotReloadHooksEnabled = true;
...
 /// All calls of [use] must be made outside of conditional checks and always in the same order.
 ///
 /// See [Hook] for more explanations.
 // ignore: deprecated_member_use, deprecated_member_use_from_same_package
 R use<R>(Hook<R> hook) => Hook.use(hook);
 
-/// [Hook] is similar to a [StatelessWidget], but is not associated
-/// to an [Element].
+/// Allows a [Widget] to create and access its own mutable data
+/// without implementing a [State].
```
nate-thegrate, line 22: I thought that "not associated to an Element" might lead to confusion, because much like a `StatelessWidget`, `Hook` has a `createState()` method, and the object returned has access to the `context`.

packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -10,20 +10,27 @@ bool debugHotReloadHooksEnabled = true;
...
+/// Whereas [Widget]s store the immutable configuration for UI components,
+/// [Hook]s store immutable configuration for any type of object.
+/// The [HookState] of a [Hook] is analogous to the [State] of a [StatefulWidget],
+/// and a single [HookWidget] can use more than one [Hook].
+///
+/// Hooks can be used by replacing `extends StatelessWidget` with `extends HookWidget`,
+/// or by replacing `Builder()` with `HookBuilder()`.
+///
+/// Hook functions must be called unconditionally during the `build()` method,
+/// and always in the same order.
```
nate-thegrate, line 33: > A [Hook] is typically the equivalent of [State]

This makes sense, since you can go by the philosophy of "rather than making a State, make a Hook instead".
However, in my mind, the equivalent of `State` would be `HookState` rather than `Hook`.
My hope is that the adjusted wording creates a good intuition while avoiding this caveat.

<br>

> A [Hook] is created within the [HookState.build] method of a [HookWidget]

I believe this sentence was a typo: a `HookState` is created by a `Hook`, not the other way around.

packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -210,13 +217,17 @@ Calling them outside of build method leads to an unstable state and is therefore
   /// HookState createState() => _MyHookState();
   /// ```
   ///
-  /// The framework can call this method multiple times over the lifetime of a [HookWidget]. For example,
+  /// The framework can call this method multiple times over the lifetime of a [HookElement]. For example,
```
nate-thegrate, line 220: This is mostly just Flutter pedantry, hopefully that's okay :)

- "Widget" = immutable, frequently thrown away and re-constructed
- "Element" = has an extended lifetime

packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -277,22 +288,19 @@ abstract class HookState<R, T extends Hook<R>> with Diagnosticable {
...
+  /// [HookElement.build] will return a cached value instead of rebuilding each [Hook].
   ///
-  /// - [setState] was called
-  /// - a previous hook's [shouldRebuild] returned `true`
-  /// - the associated [HookWidget] changed.
+  /// This method is not evaluated if a previous Hook called [markMayNeedRebuild]
+  /// and its [shouldRebuild] method returned `true`.
+  /// Additionally, if [setState], [didUpdateHook], or [HookElement.didChangeDependencies] is called,
+  /// the build is unconditional and the `shouldRebuild()` call is skipped.
   bool shouldRebuild() => true;
```
nate-thegrate, line 298: I thought the original description was really good, but I saw that (1) a rebuild will happen anyway if "a previous hook's `shouldRebuild` returned `true`" and (2) the method returns `true` by default… and for some reason I totally missed that only the hooks that called `markMayNeedRebuild` are checked.

I also saw that the 3 bullet points technically didn't cover every possibility, since an `InheritedWidget` could also trigger an unconditional rebuild.

packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -576,14 +587,12 @@ Type mismatch between hooks:
   }
 }
 
-/// A [Widget] that can use a [Hook].
+/// A [Widget] that can use [Hook]s.
```
nate-thegrate, line 590: My hope was to tweak the wording here, so it's more explicit that you can use multiple Hooks.

packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -10,20 +10,27 @@ bool debugHotReloadHooksEnabled = true;
 
 /// Registers a [Hook] and returns its value.
 ///
-/// [use] must be called within the `build` method of either [HookWidget] or [StatefulHookWidget].
+/// This function must be called inside the `build` method of a widget
+/// that uses a [HookElement] as its build context.
```
rrousselGit, line 14: I don't think that's very helpful for the average Flutter dev. They don't know what elements are usually

packages/flutter_hooks/lib/src/framework.dart
```diff
@@ -614,11 +624,17 @@ class _StatefulHookElement extends StatefulElement with HookElement {
...
 }
 
-/// Obtains the [BuildContext] of the building [HookWidget].
+/// Returns the [HookElement] that is currently running its `build()` method.
+///
+/// Throws an error if no hook widget is currently building.
+///
+/// Generally, Hook functions must be called unconditionally, in the same order.
+/// That rule does not apply to `useContext()`, however, since instead of accessing a [Hook],
+/// it merely returns the relevant [BuildContext].
```
nate-thegrate, line 633: I can see an argument for calling `useContext()` unconditionally just for consistency, but since it can make things slightly less readable, I think it's nice to include the information here.

packages/flutter_hooks/lib/src/framework.dart
nate-thegrate: That makes sense. Happy to revert!

packages/flutter_hooks/lib/src/framework.dart
rrousselGit: I'd rather not tell people to do violate hooks rules with `useContext` specifically.  
If we want to allow `useContext` to not respect those rules, then it shouldn't be named with the `useX` convention. 

IMO diognistic tools could look for `use*` inside `Widget.build` and expect all of them to follow hook rules. I'd rather not have exceptions for this ; even if technically doable.

nate-thegrate: > IMO diognistic tools could look for `use*` inside `Widget.build` and expect all of them to follow hook rules.

That's a good point. I still think it'd be nice for developers who read these docs to have access to this info, but I agree that it shouldn't be recommended. I'll try to adjust the wording here accordingly.

packages/flutter_hooks/lib/src/framework.dart
nate-thegrate: Actually, I saw that the [StatefulWidget docs](https://main-api.flutter.dev/flutter/widgets/StatefulWidget/createState.html) use the same wording, so I'll go ahead and revert here!

nate-thegrate:
Quick update—I've reverted the changes to the `useContext` description.

Hopefully now we can just focus on the uncontroversial changes here. I especially like the implementation of `markMayNeedRebuild`; I'd be happy if future devs had an easier time taking advantage of it!

PR #471, opened 2025-05-25
feat: add  `useThrottled` hooks.

ostk0069:
close. #472

## Summary by CodeRabbit

- **New Features**
  - Introduced a new hook that allows throttling the execution of callbacks, helping to limit how frequently a callback can be triggered within a specified duration.

- **Tests**
  - Added tests to verify that the new throttling hook correctly limits callback execution and behaves as expected during rapid user interactions.

### DIFF
--- a/packages/flutter_hooks/lib/src/throttled.dart
+++ b/packages/flutter_hooks/lib/src/throttled.dart
@@ -1,0 +1,42 @@
+part of 'hooks.dart';
+
+/// widget ignore updates accordingly after a specified [duration] duration.
+///
+/// Example:
+/// ```dart
+/// String userInput = ''; // Your input value
+///
+/// // Create a throttle callback
+/// final throttle = useThrottle(duration: const Duration(milliseconds: 500));
+/// // Assume a fetch method fetchData(String query) exists
+/// Button(onPressed: () => throttle(() => fetchData(userInput)));
+/// ```
+void Function(VoidCallback callback) useThrottle({
+  Duration duration = const Duration(milliseconds: 500),
+}) {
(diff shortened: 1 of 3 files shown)


packages/flutter_hooks/lib/src/throttled.dart
```diff
@@ -0,0 +1,42 @@
...
+/// ```dart
+/// String userInput = ''; // Your input value
+///
+/// // Create a throttle callback
+/// final throttle = useThrottle(duration: const Duration(milliseconds: 500));
+/// // Assume a fetch method fetchData(String query) exists
+/// Button(onPressed: () => throttle(() => fetchData(userInput)));
+/// ```
+void Function(VoidCallback callback) useThrottle({
+  Duration duration = const Duration(milliseconds: 500),
```
rrousselGit, line 15: This value seems arbitrary. I'd rather not have a default value and make the parameter required

packages/flutter_hooks/lib/src/throttled.dart
```diff
@@ -0,0 +1,42 @@
...
+/// Example:
+/// ```dart
+/// String userInput = ''; // Your input value
+///
+/// // Create a throttle callback
+/// final throttle = useThrottle(duration: const Duration(milliseconds: 500));
+/// // Assume a fetch method fetchData(String query) exists
+/// Button(onPressed: () => throttle(() => fetchData(userInput)));
+/// ```
+void Function(VoidCallback callback) useThrottle({
```
rrousselGit, line 14: Should be named `useThrottled` to match `useDebounced

